PR #20, opened 2018-01-31
general style revisions

johngruen:
@lmorchard this doesn't get us to matching with the spec but it does start down that road...

This does a few things:

1. locks down the background to match spec
2. Matches the download button to spec
3. matches the color type selector to spec
4. gets the `.doll` element closer to spec
5. generally cleans up the UI

I think with this merged it will be a lot easier to file specific bugs and figure out exactly where the current visual spec needs improvement.


src/web/index.scss
```diff
@@ -2,46 +2,18 @@
...
   --white: #fff;
+  --unit-box-shadow: 0 5px 10px 5px rgba(0, 0, 0, .1);
+  --unit-border-radius: 4px;
 }
 
 html,
 * {
   box-sizing: border-box;
   color: var(--grey-90);
+  font-family: "Fira Sans";
```
johngruen: Oh, i should revert this

lmorchard:
:+1:

Issue #19, opened 2018-01-31
provide a clean version of bg_texture_02.svg


johngruen:
The current SVG is all messed up. Please provide a cleaned up version

youwenliang:
I updated the link, should be fixed now.
https://github.com/mozilla/testpilot-assets/blob/master/ThemesRFun/00_Assets/bg_texture_02.svg

lmorchard:
dbb6cc1

PR #25, opened 2018-02-01
match doll to spec

johngruen:
Better matches doll to UX spec, fixes missing background image


johngruen:
Fixes #19

lmorchard:
> Fixes #19

FWIW, you have to put that in the actual commit comment for github to auto-close the issue

PR #32, opened 2018-02-02
add preset themes

johngruen:
This PR makes it possible to select from preset themes. It also makes the doll component flexible so that a `size` prop can control the layout.

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -43,3 +43,73 @@ export const defaultColors = {
     l: 65
   },
 };
+
+export const presetColors = [
+  {
+    toolbar: {
+      h: 0,
+      s: 50,
+      l: 50,
+      a: 100
+    },
+    toolbar_text: {
+      h: 0,
+      s: 0,
+      l: 0,
+    },
+    accentcolor: {
+      h: 0,
+      s: 0,
+      l: 0
+    },
+    textcolor: {
+      h: 0,
+      s: 0,
+      l: 0
+    },
+    toolbar_field: {
+      h: 0,
+      s: 0,
+      l: 0,
+      a: 100
+    },
+    toolbar_field_text: {
+      h: 0,
+      s: 0,
+      l: 100
+    }
+  }, {
+    toolbar: {
+      h: 180,
+      s: 50,
+      l: 50,
+      a: 100
+    },
+    toolbar_text: {
+      h: 360,
+      s: 100,
+      l: 100,
+    },
+    accentcolor: {
+      h: 360,
+      s: 100,
+      l: 100
+    },
+    textcolor: {
+      h: 360,
+      s: 100,
+      l: 100
+    },
+    toolbar_field: {
+      h: 360,
+      s: 100,
+      l: 100,
+      a: 100
+    },
+    toolbar_field_text: {
+      h: 0,
+      s: 0,
+      l: 0
+    }
+  }
+];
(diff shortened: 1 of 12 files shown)


src/lib/constants.js
```diff
@@ -43,3 +43,73 @@ export const defaultColors = {
     l: 65
   },
 };
+
+export const presetColors = [
```
johngruen: for now i just added two preset colors to demonstrate the feature. We might want to move this to another file since it could get really long.

src/web/lib/components/App/index.js
```diff
@@ -6,8 +6,10 @@ import { actions, selectors } from '../../../../lib/store';
 import AppBackground from '../AppBackground';
 import BrowserPreview from '../BrowserPreview';
 import ThemeColorsEditor from '../ThemeColorsEditor';
+import PresetThemeSelector from '../PresetThemeSelector';
 import ThemeBackgroundPicker from '../ThemeBackgroundPicker';
 import ExtensionInstallButton from '../ExtensionInstallButton';
+import ThemeUrl from '../ThemeUrl';
```
johngruen: To make the BrowserPreview component more composable, i moved the in content display of the url into it's own component. Right now, this component is stubbed out and does not accept any props.

src/web/lib/components/AppBackground/index.scss
```diff
@@ -6,6 +6,7 @@
   right: 0;
   top: 0;
   z-index: 0;
+  transition: background-image 150ms;
```
johngruen: this doesn't work....i should remove it.

src/web/lib/components/BrowserPreview/index.js
```diff
@@ -10,16 +10,37 @@ const bgImages = require.context('../../../../images/', false, /bg-.*\.png/);
...
 export const BrowserPreview = ({
+  size,
   theme,
   setSelectedColor,
-  selectedTab = 0
+  selectedTab = 0,
+  children = null,
+  setTheme = null
 }) => {
   const clickSelectColor = name => e => {
```
johngruen: @lmorchard i'm not sure if there's a better way to selectively bind click handlers. I guess i could do the logic in the JSX, but i'm not sure quite how that would look.

src/web/lib/components/BrowserPreview/index.js
```diff
@@ -63,67 +84,59 @@ export const BrowserPreview = ({
   const headerBackgroundImage = bgImages.keys().includes(theme.images.headerURL)
     ? `url(${bgImages(theme.images.headerURL)})`
     : '';
-
   return (
-    <div className="doll">
+    <div className={`doll doll--${size}`} onClick={clickDoll}>
```
johngruen: Made this component BEM b/c it's got a lot of stuff going on

src/web/lib/components/PresetThemeSelector/index.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import BrowserPreview from '../BrowserPreview';
+
+import { presetColors } from '../../../../lib/constants';
```
johngruen: Might be better to get the presetColors as props.

src/web/lib/components/ThemeUrl/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import './index.scss';
+
+// TODO: Actually show a real URL here
```
johngruen: This is just a stub for now

src/web/lib/components/PresetThemeSelector/index.js
lmorchard: Yeah, though I think this is okay for now. Maybe refactor once we want to expand the list of themes

Issue #50, opened 2018-02-07
Theme data validation / ensure good default


lmorchard:
Still kind of baffled by how the theme data ended up being bad from [chuck's screengrab](https://github.com/mozilla/ThemesRFun/pull/34#issuecomment-363450965). Did [a quick patch](https://github.com/mozilla/ThemesRFun/pull/34/files#diff-dd2928a10097fea354d05b27c5c36d87R9) that seemed to fix the error, but I think the error was unrelated to [that PR](https://github.com/mozilla/ThemesRFun/pull/34) and symptomatic of a larger issue. 

And I think that larger issue is that there's nothing to validate incoming theme JSON data or swap in a known good default if that data is bad. Should fix that with a graceful fallback

![2018-02-06 08_06_44](https://user-images.githubusercontent.com/23885/35866548-bcdadc96-0b14-11e8-9aca-78caa5fb71a4.gif)

lmorchard:
FWIW: Need to make sure validation takes place in both the add-on and the site content

PR #51, opened 2018-02-07
Tweak AppBackground to not set background style if theme colors unavailable

lmorchard:
This is a followup to [pdehaan's comment on PR 34](https://github.com/mozilla/ThemesRFun/pull/34#discussion_r166367133)


lmorchard:
Also filed #50 to dig deeper into this problem (still can't reproduce though, so 😕 )

PR #54, opened 2018-02-07
Use a transparent 1x1 image for headerURL

### LINKED ISSUE #22: background image doubles at right side of browser
Not sure exactly what's going on here but...

screen shot 2018-01-31 at 10 27 04 pm

lmorchard:
This is a hack for an upstream limitation in the theme API

We can set `images.headerURL` which gives us a *non-tiling* image
anchored to the upper right. That's where the first  of the doubles
appears.

We can set `images.additional_backgrounds` which gives us images
that have customizable alignment and tiling.

If we don't set `headerURL`, we don't get the additional backgrounds at
all. So, here's a hack to set a transparent pixel for `headerURL`

Fixes #22

### DIFF
--- a/src/extension/background.js
+++ b/src/extension/background.js
@@ -1,5 +1,8 @@
 import { makeLog, colorToCSS } from '../lib/utils';
 
+// Blank 1x1 PNG from http://png-pixel.com/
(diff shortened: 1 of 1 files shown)


chuckharmston:
Works as advertised. Total hack, but ¯\_(ツ)_/¯

PR #59, opened 2018-02-09
Add theme approval step if add-on is available

### LINKED ISSUE #27: Add theme approval step if existing user receives theme link
If an installed user receives a new theme link, they should be prompted to accept it before their browser's ui gets updated.

Spec here:
https://mozilla.github.io/testpilot-assets/ThemesRFun/01_Measurements/ThemesRFun/

lmorchard:
If ?theme param is available on initial page load, assume that this is a
shared theme. Populate the editor UI with that theme and also stash it as a
pendingTheme.

If the add-on is not installed, pendingTheme will do nothing and the
editor will display the incoming shared theme.

If the add-on is installed and responds with a stored current theme,
that theme will replace the one in the editor. Then, a permission
dialog with a preview of the pendingTheme will be presented.

If the user hits "skip" on the dialog, pendingTheme is discarded.

If the user hits "apply", the pending theme is pushed into the add-on
and replaces the theme in the editor.

Fixes #27


lmorchard:
Okay, this one could use some help or at least some sanity checking:

First thing is that the theme preview in the dialog is not quite right. Still wrestling with the CSS for the doll a little, but would totally accept some help.

Second thing is that there's a flash of theme changing on startup that seems hard to work around. There's different behavior depending on whether the add-on's installed or not. But, we can't ever really know whether the add-on is *not* installed - we only know whether it responded to a message. So, not getting a response means either it's not installed or it's just taking awhile.

So, what happens is that the uninstalled state appears first, and then everything switches to the installed state. That means the editor switches between themes just as the dialog pops up. Also the CTA to install the add-on goes away.

We might be able to paper this over with a loading spinner, but the only thing we can really do there is just arbitrarily wait a second or two.

src/web/index.js
```diff
@@ -115,7 +124,17 @@ const params = queryString.parse(window.location.search);
 if (!params.theme) {
   postMessage('fetchTheme');
 } else {
-  jsonCodec.decompress(params.theme).then(theme => {
-    store.dispatch(actions.theme.setTheme({ theme }));
+  log('Received shared theme');
+  // TODO: figure out if this param theme matches theme from add-on and ignore if so
+  urlDecodeTheme(params.theme).then(theme => {
+    // Set the pending theme - only matters if add-on is installed
```
lmorchard: This is where the complicated part happens...

src/web/index.js
```diff
@@ -115,7 +124,17 @@ const params = queryString.parse(window.location.search);
 if (!params.theme) {
   postMessage('fetchTheme');
 } else {
-  jsonCodec.decompress(params.theme).then(theme => {
-    store.dispatch(actions.theme.setTheme({ theme }));
+  log('Received shared theme');
+  // TODO: figure out if this param theme matches theme from add-on and ignore if so
```
lmorchard: Oh also TBD: If the ?theme param matches what the current theme already is, we should skip displaying the permission dialog. Haven't implemented that check yet.

lmorchard:
Hmm, and actually it looks like this flashes the shared theme in the browser temporarily too. That seems like a bug in the probably over-zealous Redux middleware. Will poke at that

lmorchard:
Debugged middleware a bit, think I fixed the bug with accidentally applying the shared theme to the browser. Still thinking through the sequence & race conditions on initial load, though

chuckharmston:
Proposal: we review this PR for its own merits, and defer consideration of the sequencing pending UX on #62.

src/lib/utils.js
```diff
@@ -10,3 +10,8 @@ export const colorToCSS = color => {
     ? `hsl(${h}, ${s}%, ${l}%)`
     : `hsla(${h}, ${s}%, ${l}%, ${color.a * 0.01})`;
 };
+
+export const themesEqual = (themeA, themeB) =>
```
lmorchard: Not using this yet, which reminds me of the other thing I'm trying to figure out: When to decide that the shared theme matches the current theme.

I tried doing it as kind of a validation step - the pendingTheme state would not accept a value that matches the current theme. But that ran into race conditions in add-on messaging. Could just check on an arbitrary delay and auto-dismiss the accept dialog, but then there's a flash of content

chuckharmston: This seems like a good usecase for a selector; just have a boolean prop passed to the `<App>` that checks the `pendingTheme` and `currentTheme` for equality. Then it's just a matter of checking that prop, as passed through the component hierarchy.

lmorchard: I also tried this as a selector. Problem was, as soon as you change any colors, suddenly the pending theme no longer matches the current theme and the approval dialog appears. Been trying to figure out at what point to just ignore / delete the incoming pendingTheme so that doesn't happen

lmorchard:
Okay, tweaked things to add a `userHasEdited` flag to use as part of the decision as to whether the shared theme dialog should be displayed. Seems to help solve the issue of skipping the dialog if the shared theme matches current theme

lmorchard:
One more quick tweak for clarity & to pass linting and I think I'll leave this alone for a bit

src/lib/utils.js
```diff
@@ -4,9 +4,16 @@ export const makeLog = context => (...args) =>
   // eslint-disable-next-line no-console
   DEBUG && console.log(`[ThemesRFun ${context}]`, ...args);
 
+export const floorHSLA = color => ({
```
lmorchard: Planning to revisit this in #50 - the color picker and theme schema use different magnitudes for HSLA values, but the JS floating point math causes the values to drift. And thus, themes that should be identical look non-identical to the shouldOfferPendingTheme selector because insignificant digits have changed.

src/web/index.js
```diff
@@ -113,9 +125,22 @@ render(
 
 const params = queryString.parse(window.location.search);
```
lmorchard: Might drop a long comment here about the startup process and shared theme handling

chuckharmston:
> I tried to describe it in the commit message - is that what you're looking for / should that live somewhere else?

Yep! I'd like that type of thing more formally represented in in-repo documentation, with a bit of the "why" attached to the "what".

src/web/index.js
```diff
@@ -111,11 +123,51 @@ render(
   document.getElementById('root')
 );
 
+/**
```
lmorchard: Added this comment for now, will file an issue for further doco

lmorchard:
Also added a commit to bump the add-on version and plan to do another development branch deployment after merge

PR #69, opened 2018-02-14
Add a loading overlay while the app initializes

### LINKED ISSUE #62: Make a loader component
Let's make the page look pretty from the jump by having a cool loader component until we can determine if the user has the add-on available.

lmorchard:
- Add AppLoadingIndicator component that overlays app on startup

- Implements a timer that resets with every Redux store change. If no
  changes happen and the timer expires, we consider the app loaded -
  assuming that no more state changes means it's done loading.

- If there's no shared ?theme parameter, we start a timer immediately
  after asking for current theme from add-on. If the add-on never
  answers, the app considers itself loaded after the delay expires.

- If there's a shared ?theme parameter, we decode the parameter and the
  first state change starts the timer - thus giving things time to at
  least handle the shared theme.

Fixes #62

### DIFF
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -21,8 +21,10 @@ const clipboard = new Clipboard('.clipboardButton');
 
 const addonUrl = process.env.ADDON_URL;
 
+// Period after which app loading indicator will disappear if add-on not found
+const LOADER_DELAY_PERIOD = 500;
 const PING_PERIOD = 1000;
-const MAX_OUTSTANDING_PINGS = 2;
+const MAX_OUTSTANDING_PINGS = 3;
 let outstandingPings = 0;
 
 const jsonCodec = JsonUrl('lzma');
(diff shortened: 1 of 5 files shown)


lmorchard:
This is probably over-engineered. But, just using a simple timer always ended up being too long or too short. The timer I ended up with just kind of bounces along until state changes settle down and the app stabilizes.

src/web/index.js
```diff
@@ -123,6 +125,36 @@ render(
   document.getElementById('root')
 );
 
+/**
+ * We display a loading indicator on startup that overlays the rest of the app.
```
lmorchard: This is probably over-engineered. But, just using a simple timer always ended up being too long or too short. The timer I ended up with just kind of bounces along until state changes settle down and the app stabilizes.

lmorchard: But, totally willing to yank this chunk out and just drop in a straight timeout for the loader. Figured we could iterate on it along with however we want to finesse the loader component itself.

chuckharmston: I agree that this is hacky and overengineered. Can we prospectively be more precise than just hoping the state stops changing? Perhaps looking for specific values to be non-`null`?

chuckharmston:
One bug introduced by this PR: if there is a `theme` querystring param for a theme equal to the browser's current theme, you are still prompted to install the custom theme:

screen shot 2018-02-15 at 10 04 08

This does not happen on master. Screenshot was taken with `theme=XQAAAALYAAAAAAAAAABBKYkm0ypxtqu9mB2KK4d4ccIgevtJaJcbAYzNX0U8l4L11DGso36hFn2wII3OG50e6ecOWJd78ulKUYiKxYb27GorFor8QjMoI-AYfHD-W-Rnt0qVC-YziOctYEYJi9iBHFPNT4f6RAP6_aK6RiuefRpCwZM7b3D2ChzeJYYGWDQB_uH__JP4AA`

lmorchard:
> Can we prospectively be more precise than just hoping the state stops changing? Perhaps looking for specific values to be non-null?

@chuckharmston Any specific suggestions? Tried looking for some particular cues before landing on this, but couldn't come up with any. All comes down to the unpredictable timing of whether or not we have an add-on available.

lmorchard:
> One bug introduced by this PR: if there is a theme querystring param for a theme equal to the browser's current theme, you are still prompted to install the custom theme:

Also, I'm not seeing this happen on my end. Tried it with that theme and it worked as expected without the dialog after I'd applied the theme. 

If you can reliably repro, could you maybe peek at the state of the Redux store with dev tools and see if the pending theme and current theme are *exactly* the same - or if they're slightly off in floating point digits? There was an issue I was hacking around with otherwise identical themes getting screwed up by JS math

chuckharmston:
It was, in fact, floating point math. Great call. I'll file a separate bug to address that.

PR #71, opened 2018-02-16
Enforce normal structure for internal theme data

### LINKED ISSUE #50: Theme data validation / ensure good default

lmorchard:
Add filter functions to ensure we have conistent theme data. This
should ensure that whatever gets passed in as a shared ?theme param is
handled gracefully, and that arbitrary data isn't accepted by the
add-on.

Fixes #50

### DIFF
--- a/src/extension/background.js
+++ b/src/extension/background.js
@@ -1,4 +1,4 @@
-import { makeLog, colorToCSS } from '../lib/utils';
+import { makeLog, colorToCSS, normalizeTheme } from '../lib/utils';
 
 // Blank 1x1 PNG from http://png-pixel.com/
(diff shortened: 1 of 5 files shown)


lmorchard:
Small tweak to make the state filtering more granular

PR #74, opened 2018-02-20
Allow undo/redo to push theme changes to browser

lmorchard:
Turns out that the theme in the browser wasn't changing on undo/redo - this should fix

### DIFF
--- a/src/lib/store.js
+++ b/src/lib/store.js
@@ -1,6 +1,6 @@
 import { createStore, combineReducers } from 'redux';
 import { createActions, handleActions } from 'redux-actions';
-import undoable, { ActionCreators } from 'redux-undo';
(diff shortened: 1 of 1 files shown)


src/lib/store.js
```diff
@@ -1,14 +1,20 @@
...
 import {
   normalizeTheme,
   normalizeThemeColor,
   normalizeThemeColors
 } from './utils';
 
 // Actions that should trigger a theme update in URL history and the add-on
-export const themeChangeActions = ['SET_THEME', 'SET_COLOR', 'SET_BACKGROUND'];
+export const themeChangeActions = [
+  'SET_THEME',
```
chuckharmston: A nit: since it's a string with meaningful content, I'd prefer to see this as a constant.

Issue #43, opened 2018-02-06
Implement saving themes to local storage with previews


lmorchard:
From [acceptance criteria](https://github.com/mozilla/ThemesRFun/blob/master/docs/acceptance.md):

> - [ ] It should let users save themes they've generated to localStorage and see previews of them on subsequent visits.

johngruen:
@youwenliang I think we still need UX for this...Don't worry about a pixel perfect visual spec

PR #75, opened 2018-02-21
Implement backend for saving themes to localStorage with previews

lmorchard:
UX is pending for this feature, but we can at least get the storage
mechanics working.

- Add ui.savedThemes state with associated actions, reducer, & selector

- Add storage module to handle localStorage manipulation for themes, as
  well as updating Redux state on changes

- Provisional ThemeSaveButton and SavedThemeSelector components

- DRY up selectors passed into App

Issue #43

### DIFF
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -17,6 +17,7 @@ import {
   themeChangeActions
 } from '../lib/store';
 import App from './lib/components/App';
+import storage from './lib/storage';
 
 import './index.scss';
 
@@ -78,6 +79,8 @@ const store = createAppStore(
   )
 );
 
+storage.init(store);
+
 window.addEventListener('popstate', ({ state: { theme } }) =>
   store.dispatch({
     ...actions.theme.setTheme({ theme }),
@@ -122,7 +125,12 @@ setInterval(() => {
 
 render(
   <Provider store={store}>
-    <App {...{ addonUrl, urlEncodeTheme, clipboard }} />
+    <App {...{
+      addonUrl,
+      urlEncodeTheme,
+      clipboard,
+      storage
+    }} />
   </Provider>,
   document.getElementById('root')
 );
--- a/src/web/lib/components/App/index.js
+++ b/src/web/lib/components/App/index.js
@@ -12,27 +12,29 @@ import ExtensionInstallButton from '../ExtensionInstallButton';
 import SharedThemeDialog from '../SharedThemeDialog';
 import AppLoadingIndicator from '../AppLoadingIndicator';
 import ThemeUrl from '../ThemeUrl';
+import ThemeSaveButton from '../ThemeSaveButton';
+import SavedThemeSelector from '../SavedThemeSelector';
 
 import './index.scss';
 
-const mapStateToProps = state => ({
-  theme: selectors.theme(state),
-  themeCanUndo: selectors.themeCanUndo(state),
-  themeCanRedo: selectors.themeCanRedo(state),
-  hasExtension: selectors.hasExtension(state),
-  loaderDelayExpired: selectors.loaderDelayExpired(state),
-  selectedColor: selectors.selectedColor(state),
-  shouldOfferPendingTheme: selectors.shouldOfferPendingTheme(state),
-  pendingTheme: selectors.pendingTheme(state)
-});
+const mapStateToProps = state => {
+  const mappedSelectors = Object.entries(selectors).reduce(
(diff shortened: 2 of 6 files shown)


src/web/lib/components/App/index.js
```diff
@@ -12,27 +12,30 @@ import ExtensionInstallButton from '../ExtensionInstallButton';
...
-  themeCanUndo: selectors.themeCanUndo(state),
-  themeCanRedo: selectors.themeCanRedo(state),
-  hasExtension: selectors.hasExtension(state),
-  loaderDelayExpired: selectors.loaderDelayExpired(state),
-  selectedColor: selectors.selectedColor(state),
-  shouldOfferPendingTheme: selectors.shouldOfferPendingTheme(state),
-  pendingTheme: selectors.pendingTheme(state)
-});
+const mapStateToProps = state => {
+  const mappedSelectors = {};
```
lmorchard: This might be a too-clever idea, but looked like the pattern here was to just keep adding straight mappings of all selectors to the state.

src/web/lib/storage.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const generateThemeKey = () =>
+  `${Date.now()}-${Math.floor(Math.random() * 1000)}`;
+
+export function putTheme(key, theme) {
+  log('putTheme', key, theme);
+  const storageKey = themeStorageKey(key);
+  localStorage.setItem(
+    storageKey,
+    JSON.stringify({
```
lmorchard: FWIW, trying to leave room for expansion here - e.g. if we want to store a name / description along with the theme.

src/web/lib/components/ThemeSaveButton/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+import './index.scss';
+
+export const ThemeSaveButton = ({
+  theme,
+  savedThemes,
+  generateThemeKey,
+  putTheme
+}) => {
+  const saveTheme = () => putTheme(generateThemeKey(), theme);
```
lmorchard: If we want to use user-supplied names in the UX, we can switch from this randomly generated key to something from a text field.

src/web/lib/storage.js
```diff
@@ -0,0 +1,89 @@
...
+      const theme = getTheme(themeKey);
+      if (theme !== null) {
+        themes[themeKey] = theme;
+      }
+    }
+  }
+  return themes;
+}
+
+// HACK: Storage events are dispatched to all windows *except* the one that
```
lmorchard: Gross, but 🤷‍♂️ it keeps the event handling consistent

lmorchard:
@johngruen You might want to play with this stuff with an eye for UX. I just straight punted on styling for now so we can iterate

src/web/lib/components/SavedThemeSelector/index.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import BrowserPreview from '../BrowserPreview';
+
+import './index.scss';
+
+export const SavedThemeSelector = ({ setTheme, savedThemes, deleteTheme }) => {
+  const sortedSavedThemes = Object.entries(savedThemes).sort(
```
lmorchard: Could move this to a selector in store.js, but punted for now to make it easier to play with the presentation

johngruen:
@lmorchard will check this out in a bit

src/web/lib/components/SavedThemeSelector/index.js
chuckharmston: Fair.

src/web/lib/storage.js
chuckharmston: ¯\\_(ツ)_/¯

src/web/index.js
```diff
@@ -17,6 +17,7 @@ import {
   themeChangeActions
 } from '../lib/store';
 import App from './lib/components/App';
+import * as storage from './lib/storage';
```
chuckharmston: Might just be the Pythonista in me, but `*` imports kind of stink. Any chance you could `export default { init, etc }` from `lib/storage` instead?

lmorchard: I kind of don't like the repetition in listing all the exports a second time, but I also don't super like the `*` import either. I'll go with the latter though and tweak it

lmorchard:
Tweaked `mappedSelectors` to use `reduce()` and dropped the `*` import. Will merge when circleci goes green!

PR #78, opened 2018-02-23
add footer

johngruen:
Adds the footer. The /legal links will have to be addressed in a later PR...maybe they could trigger a modal or something? Send's version of the same looks like this: https://send.firefox.com/legal

Fixes #78


lmorchard:
> The /legal links will have to be addressed in a later PR...maybe they could trigger a modal or something?

Can you file an issue with exactly what this should be, so we don't lose track of it?

PR #85, opened 2018-02-23
Add stylelint support for extra fresh Sass

pdehaan:


src/web/lib/components/Banner/index.scss
```diff
@@ -12,8 +11,7 @@ $banner-height: 71px;
   height: $banner-height;
   position: fixed;
   width: 100%;
-  box-shadow: 0 0 0 1px rgba(0, 0, 0, .05),
-              0 0 10px rgba(0, 0, 0, .05);
+  box-shadow: 0 0 0 1px rgba(0, 0, 0, .05), 0 0 10px rgba(0, 0, 0, .05);
```
pdehaan: I think Sass supports named colors and hex colors for `rgb()` and `rgba()` so we _might_ be able to tighten this up slightly by going with:

```sass
box-shadow: 0 0 0 1px rgba(#000, .05), 0 0 10px rgba(#000, .05);
```

src/web/lib/components/ThemeColorsEditor/index.scss
```diff
@@ -17,13 +17,15 @@
...
       }
+
       &:hover {
-        background: rgba(12, 12, 13, 0.1);
+        background: rgba(12, 12, 13, .1);
       }
+
       &:active {
-        background: rgba(12, 12, 13, 0.2);
+        background: rgba(12, 12, 13, .2);
```
pdehaan: We seem to use this `rgba(12, 12, 13, n)` color quite a bit (6-7 places at least). Not sure if it's worth extracting out into some Sass variable so we could do `rgba($light-blue, .2)` or something. Possibly/probably not worth the effort though.

src/web/lib/components/Banner/index.scss
```diff
@@ -3,7 +3,6 @@
 $banner-height: 71px;
 
 .banner__content {
-  position: fixed;
```
pdehaan: This was deleted since it was specified below on L13.

src/web/lib/components/AppHeader/index.scss
```diff
@@ -47,10 +47,9 @@
     position: absolute;
     right: 0;
     text-decoration: none;
-    transition: size 125ms;
-    user-select: none;
```
pdehaan: This was deleted since it was duplicated on L55 below.

src/web/lib/components/AppHeader/index.scss
```diff
@@ -47,10 +47,9 @@
     position: absolute;
     right: 0;
     text-decoration: none;
-    transition: size 125ms;
```
pdehaan: I moved this down a couple lines since it seems like we're also transitioning the `width` property, so I grouped them together. ¯\\\_(ツ)\_/¯

johngruen:
@pdehaan i just did a pretty major refactor on CSS. I'm happy to merge the linting rules, but maybe not the changes.

pdehaan:
Rebased.
Not sure you can just take the linting rules without the rest of the changes, since the changes were to fix CSS warnings/errors.
If you're thinking the CSS linting is too strict, we could extend the "stylelint-config-recommended" ruleset instead of the more strict/complete "stylelint-config-standard" ruleset.

PR #86, opened 2018-02-24
Tweak a bunch of a11y settings from aXe

pdehaan:
Addresses a bunch of results from aXe, except for ~20 color contrast issues.
Will probably want to run this through a few other a11y linters to make sure I didn't miss some stuff.

### Current status:

lighthousereport

...

lighthousereport2

That "Performance" number is probably v.misleading since I'm running locally on probably an `process.env.NODE_ENV=development` build.

### DIFF
--- a/src/web/index.html.ejs
+++ b/src/web/index.html.ejs
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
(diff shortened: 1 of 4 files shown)


johngruen:
This looks great...i can rebase this into my changes once they land.

pdehaan:
Rebased, if you want it.

Issue #37, opened 2018-02-06
Consider passing color values in URL rather than hash


chuckharmston:
That'd make it a lot easier to do colorspace analysis.

lmorchard:
Can you elaborate on what you mean? The theme is currently a compressed JSON object encoded as base64 in the URL param - it's not using the hash

lmorchard:
Oh, you mean serialize the JSON as set of URL parameters? (By hash, I thought you meant the fragment identifier #hash) I tried doing that at one point, but the URL handling got kind of gnarly. We can take another swing at it, though. This is the kind of data we'll have to serialize/deserialize, maybe need to work on munging property names a bit to reduce length:

```json
{
 "images": "",
 "colors": {
  "toolbar": {
   "h": 0,
   "s": 50,
   "l": 50,
   "a": 100
  },
  "toolbar_text": {
   "h": 0,
   "s": 0,
   "l": 0
  },
  "accentcolor": {
   "h": 0,
   "s": 0,
   "l": 0
  },
  "textcolor": {
   "h": 0,
   "s": 0,
   "l": 0
  },
  "toolbar_field": {
   "h": 0,
   "s": 0,
   "l": 0,
   "a": 100
  },
  "toolbar_field_text": {
   "h": 0,
   "s": 0,
   "l": 100
  }
 }
}
```

lmorchard:
FWIW, alternatively, we could ping some metrics service with the de-serialized theme data directly

johngruen:
Could we do something like trf.com/?a=ffccaa&b=ccaabb&c=aaffaa...?

lmorchard:
I think the format would depend on exactly how and where we want to do the color metrics. 

And also how often do we want to record it - because the URL changes for every change the user makes to the theme. Would it be better to record the theme at more significant points? (e.g. when sharing out or receiving a shared theme?)

johngruen:
I added these as custom dimensions using HSL(A) values passed in as CSVs. The idea would be that this format would be the easiest to parse in a big SQL table, but it might be no different to send them as HEX or HEXa

lmorchard:
So, does that look at the URL (and so we change the URL format) or do we ping it from JS?

lmorchard:
But, I'm saying we might not even need to tweak the URL if we use something else to measure this (i.e. pings from JS with the measurement protocol)

lmorchard:
[...]

Issue #28, opened 2018-02-01
Provide Firefox Download link if user is in a different browser


johngruen:
If a user visits TRF from a non-firefox browser, we should prompt them to download firefox:

Spec is here: https://mozilla.github.io/testpilot-assets/ThemesRFun/01_Measurements/ThemesRFun/#artboard2

Note: do not implement the dismiss button

URL should be something like this with the final name slotted in:
```
https://www.mozilla.org/firefox/new/?utm_campaign=${tbd}-acquisition&utm_medium=referral&utm_source=${tbd}.firefox.com
```

Issue #11, opened 2018-01-23
Add Metrics Schema


johngruen:

lmorchard:
Tweaked the title on this - we *have* a metrics schema, we need to finish whatever tweaks are necessary

lmorchard:
A thing that occurred to me - we have this ping in metrics.md:
```
ec: theme-change,
ea: select-full,
ev: {$id} // unique integer value id of the selected theme
```
That `{$id}` should probably change to a title or a filename of the theme JSON. Currently, the "id" in the code is the index of the preset theme in the list. But, that will be unstable and change based on the ordering of those themes - which right now is just filename alpha sort.

apoorvasingh17:
I would like to work on this issue. Can someone give me pointers?

pdehaan:
@apoorvasingh17 probably best to start on a “good first issue”; https://github.com/mozilla/Themer/labels/good%20first%20issue

lmorchard:
Do we have any further changes to make here for mvp versus what is already in metrics.md? /cc @johngruen

johngruen:
@lmorchard will check this today

johngruen:
Making some notes here...
- [ ] Doesn't look like the `tab_line` has a value. this makes sense since we added it later.
- [ ] in `src/lib/metrics` comments say we're passing hsla colors, but they're rgb(a), we should correct that.

lmorchard:
>  in src/lib/metrics comments say we're passing hsla colors, but they're rgb(a), we should correct that.

Would be easiest to change metrics.md than convert all the internal RGBA values to HSLA on every metric hit, unless there's a reason they absolutely have to be HSLA

johngruen:
@lmorchard yeah, i'll add that to the metrics docs PR

PR #89, opened 2018-02-26
Implement current set of metrics pings

### LINKED ISSUE #46: Implement GA metrics pings
From [acceptance criteria](https://github.com/mozilla/ThemesRFun/blob/master/docs/acceptance.md):

> - [ ] It should ping GA for all pings described in the [experiment metrics doc](./metrics.md).
> - [ ] It should respect DNT.

lmorchard:
- New Metrics module with methods to send various pings

- Add a random per-client UUID to localStorage, generating as necessary

- Wire up metrics pings to links and major app interactions

- Implement timer-based add-on install success/fail event

- Simplify BrowserPreview to only handle onClick rather than setting the
  theme directly, so that presets and saved themes can handle metric
  pings differently

- Only deselect current color if there's one selected

Fixes #46

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,9 @@
   ],
   "private": true,
   "main": "src/web/index.js",
+  "config": {
+    "GA_TRACKING_ID": "UA-114768519-1"
+  },
   "scripts": {
     "start": "npm-run-all --parallel server watch:extension watch:lint",
     "server": "cross-env NODE_ENV=development webpack-dev-server --config webpack.web.js",
@@ -83,7 +86,8 @@
     "redux": "^3.7.2",
     "redux-actions": "^2.2.1",
     "redux-devtools-extension": "^2.13.2",
-    "redux-undo": "^1.0.0-beta9-9-7"
+    "redux-undo": "^1.0.0-beta9-9-7",
+    "testpilot-ga": "^0.3.0"
   },
   "extensionManifest": {
     "name": "ThemesRFun",
--- a/src/lib/metrics.js
+++ b/src/lib/metrics.js
@@ -1,0 +1,278 @@
+import TestPilotGA from 'testpilot-ga';
+
+// Use of package.json for configuration
+import packageMeta from '../../package.json';
+
+import { DEBUG, makeLog } from './utils';
+import { selectors } from './store';
+import { colorsWithAlpha } from './constants';
+
+const log = makeLog('metrics');
+
+let analytics = null;
+
+// Time after which to consider add-on install to have failed, because we have
+// no good way to tell if the install has been cancelled - only if the add-on
+// becomes available
+const INSTALL_FAILURE_DELAY = 10 * 1000;
+let installTimer = null;
+
+// See also: https://github.com/mozilla/ThemesRFun/blob/master/docs/metrics.md
+
+// Custom metrics
+let cm1; // integer count of select-full events from start of visit
+let cm2; // integer count of select-bg events from start of visit
+let cm3; // integer count of select-color events from start of visit
+
+// Custom dimensions
+let cd1; // does the user have the add-on installed. One of true or false
+let cd2; // did this visit originate from an add-on click. One of true or false
+let cd3; // did the user receive a theme as a query parameter. One of true or
+// false
+let cd4; // engaged with any theme-change event. One of true or false based on
+// whether user has fired any theme-change during their visit.
+let cd5; // hsla (csv) of the toolbar
+let cd6; // hsl (csv) of the toolbar_text
+let cd7; // hsl (csv) of the accentcolor
+let cd8; // hsl (csv) of the textcolor
+let cd9; // hsla (csv) of the toolbar_field
+let cd10; // hsl (csv) of the toolbar_field_text
+let cd11; // unique integer id of the background pattern selected
+
+const COLORS_TO_DIMENSIONS = {
+  toolbar: 'cd5',
+  toolbar_text: 'cd6',
+  accentcolor: 'cd7',
+  textcolor: 'cd8',
+  toolbar_field: 'cd9',
+  toolbar_field_text: 'cd10'
+};
+
+const hslaToCSV = (name, { h, s, l, a }) =>
+  `${h},${s},${l}${colorsWithAlpha.includes(name) ? `,${a}` : ''}`;
+
+const Metrics = {
+  init(appContext = 'web') {
+    /* eslint-disable prefer-destructuring */
+    analytics = new TestPilotGA({
+      aid: packageMeta.extensionManifest.applications.gecko.id,
+      an: packageMeta.extensionManifest.name,
+      av: packageMeta.version,
+      ds: appContext,
+      tid: packageMeta.config.GA_TRACKING_ID,
+      // TODO: add some env vars or use window.location to determine local / dev / stage / prod?
+      cd19: DEBUG ? 'dev' : 'production'
+    });
+    /* eslint-enable prefer-destructuring */
+
+    cm1 = 0;
+    cm2 = 0;
+    cm3 = 0;
+    cd1 = false;
+    cd2 = false;
+    cd3 = false;
+    cd4 = false;
+  },
+
+  sendEvent(...params) {
+    if (analytics) {
+      analytics.sendEvent(...params);
+    }
+  },
+
+  storeMiddleware() {
+    return ({ getState }) => next => action => {
+      const result = next(action);
+
+      if (action.type === 'SET_THEME') {
+        const theme = selectors.theme(getState());
+        this.setTheme(theme);
+        if (action.meta && action.meta.userEdit) {
+          this.setThemeChanged(true);
+        }
+      }
+
+      return result;
+    };
+  },
+
+  setHasAddon(value) {
+    cd1 = value;
+  },
+
+  setWasAddonClick(value) {
+    cd2 = value;
+  },
+
+  setReceivedTheme(value) {
+    cd3 = value;
+  },
+
+  setThemeChanged(value) {
+    cd4 = value;
+  },
+
+  themeToDimensions(theme) {
+    return Object.entries(theme.colors).reduce(
+      (acc, [name, hsla]) => ({
+        ...acc,
+        [COLORS_TO_DIMENSIONS[name]]: hslaToCSV(name, hsla)
+      }),
+      { cd11: theme.images.headerURL }
+    );
+  },
+
+  setTheme(theme) {
+    const themeDimensions = this.themeToDimensions(theme);
+    log('update theme', themeDimensions);
+    ({ cd5, cd6, cd7, cd8, cd9, cd10, cd11 } = themeDimensions);
+  },
+
+  installStart() {
+    if (installTimer) {
+      clearTimeout(installTimer);
+    }
+    installTimer = setTimeout(() => {
+      this.installFailure();
+    }, INSTALL_FAILURE_DELAY);
+
+    this.sendEvent('install-addon', 'button-click', {
+      el: 'install-trigger',
+      cm1,
+      cm2,
+      cm3,
+      cd3,
+      cd4,
+      cd5,
+      cd6,
+      cd7,
+      cd8,
+      cd9,
+      cd10,
+      cd11
+    });
+  },
+
+  installFailure() {
+    this.sendEvent('install-addon', 'poll-event', {
+      el: 'install-fail'
+    });
+  },
+
+  installSuccess() {
+    if (!installTimer) {
+      // Skip this event if there's no install timer in progress.
+      return;
+    }
+    clearTimeout(installTimer);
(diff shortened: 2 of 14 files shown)


.eslintrc.yaml
```diff
@@ -27,4 +27,5 @@ rules:
   no-restricted-syntax: off
   no-use-before-define: off
   object-curly-newline: off
+  prefer-destructuring: off
```
lmorchard: Kept getting a lint error for this rule in metrics.js with no line number. Spent 20 minutes trying to figure out what it was talking about, then gave up and did this.

package.json
```diff
@@ -8,6 +8,9 @@
   ],
   "private": true,
   "main": "src/web/index.js",
+  "config": {
+    "GA_TRACKING_ID": "UA-114768519-1"
```
lmorchard: This is a temporary ID for a dashboard I made on my account. We should change it for production and possibly make it switch between IDs for prod / dev builds.

src/lib/metrics.js
```diff
@@ -0,0 +1,334 @@
...
+
+// Time after which to consider add-on install to have failed, because we have
+// no good way to tell if the install has been cancelled - only if the add-on
+// becomes available
+const INSTALL_FAILURE_DELAY = 10 * 1000;
+let installTimer = null;
+
+// See also: https://github.com/mozilla/ThemesRFun/blob/master/docs/metrics.md
+
+// Custom metrics
```
lmorchard: Some of these comments for dimensions are messy, because I copypasta'd from metrics.md and turned them into JS with a regex. Might clean them up, but figured there might be changes to metrics.md yet (#11)

src/lib/metrics.js
```diff
@@ -0,0 +1,334 @@
...
+      cd9,
+      cd10,
+      cd11
+    });
+  },
+
+  linkClick(el) {
+    // if el === download-firefox, add the following dimensions to this event
+    const downloadFirefoxDimensions =
+      el !== 'download-firefox' ? {} : { cd5, cd6, cd7, cd8, cd9, cd10, cd11 };
```
lmorchard: We don't actually have this link anywhere, yet. (#28)

src/web/index.js
```diff
@@ -99,6 +109,8 @@ window.addEventListener('message', ({ source, data: message }) => {
       const hasExtension = selectors.hasExtension(store.getState());
       if (!hasExtension) {
         store.dispatch(actions.ui.setHasExtension({ hasExtension: true }));
+        Metrics.installSuccess();
+        postMessage('setClientUUID', { clientUUID: Metrics.getClientUUID() });
```
lmorchard: This does nothing, yet. Still scratching my head on implementing [the `finish-visit` event](https://github.com/mozilla/ThemesRFun/blob/master/docs/metrics.md#finish-visit), which I imagine will require the add-on to make a metric ping and will need the clientUUID

src/web/lib/components/SavedThemeSelector/index.js
```diff
@@ -16,10 +16,20 @@ export const SavedThemeSelector = ({ setTheme, savedThemes, deleteTheme }) => {
...
+            title="delete"
+          >
             <img src={iconClose} />
           </button>
-          <BrowserPreview {...{ size: 'small', theme, setTheme }} />
+          <BrowserPreview
+            {...{
+              size: 'small',
+              theme,
+              onClick: () => setTheme({ theme })
```
lmorchard: We could probably also do `Metrics.themeChangeFull(themeId)` here, but there wasn't a specific event for selecting a saved theme. Maybe we want to add one in wrapping up #11?

lmorchard:
I think this also takes care of #37 because the theme colors get sent as custom dimensions

johngruen:
Hey @lmorchard, this is a lot to chew on! @chuckharmston could you give this a look since I'm at a work week RN and have limited availability

package.json
pdehaan: Should we file a new issue to change this before release? Sadly it's in package.json so we can't add a `// TODO` comment.

src/lib/metrics.js
```diff
@@ -0,0 +1,334 @@
...
+        method: 'POST',
+        mode: 'cors',
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8'
+        },
+        body: encoded
+      })
+        .then(() => log(`Sent GA message via fetch: ${encoded}`))
+        .catch(err => log(`GA sending via fetch failed: ${err}`));
+    }
```
pdehaan: Do we need a fallback if [`fetch()` is not supported](https://developer.mozilla.org/docs/Web/API/Fetch_API#Browser_compatibility)?

src/lib/metrics.js
```diff
@@ -0,0 +1,334 @@
...
+      cd7,
+      cd8,
+      cd9,
+      cd10,
+      cd11
+    });
+  },
+
+  sendEvent(params) {
+    if (!GA_TRACKING_ID || !clientUUID || navigator.doNotTrack === 1) {
```
pdehaan: We check `navigator.doNotTrack` here, but below on L317, we do an explicit check if `typeof navigator !== 'undefined'`. Do we need to check if `navigator` and `navigator.doNotTrack` are kosher here before checking the value?

src/web/lib/components/AppFooter/index.js
```diff
@@ -1,41 +1,77 @@
...
+export const AppFooter = () => (
   <footer className="app-footer">
     <div className="app-footer__legal">
-      <a className="app-footer__legal-link" href="https://www.mozilla.org">
-        <ReactSVG path={ iconMoz } className="app-footer__legal-logo"/>
+      <a
+        className="app-footer__legal-link"
+        onClick={() => Metrics.linkClick('mozilla-logo')}
+        href="https://www.mozilla.org"
+      >
```
pdehaan: Do we need `<a href="..." rel="noreferrer noopener">...</a>`, or is that only for `target="_blank"` links?
Or I guess, should we be adding `target="_blank"` to these links? (We have them in src/web/lib/components/AppHeader/index.js below)

src/web/lib/storage.js
```diff
@@ -1,8 +1,12 @@
+import uuidV4 from 'uuid/v4';
+
 import { actions } from '../../lib/store';
 import { makeLog } from '../../lib/utils';
 
 const log = makeLog('web.storage');
 
+const CLIENT_UUID_KEY = 'ThemesRFun-clientUUID';
```
pdehaan: Not sure if we have the final name yet, but should this be "ThemesRFun" or "Themer"?

.eslintrc.yaml
pdehaan:
```sh
$ npm run lint

> themesrfun@0.0.22 lint /Users/pdehaan/dev/github/mozilla/ThemesRFun
> eslint --color .

/Users/pdehaan/dev/github/mozilla/ThemesRFun/src/lib/metrics.js
  11:7  error  Use object destructuring  prefer-destructuring
```

It's choking on [/src/lib/metrics.js:11](https://github.com/mozilla/ThemesRFun/pull/89/files#diff-d0154faa85aed1ddcb0a1d33f0484caeR11):

```js
const GA_TRACKING_ID = packageMeta.config.GA_TRACKING_ID;
```

Solution is a bit hacky, but I couldn't figure out how to destructure nested properties on an `import` to get the `GA_TRACKING_ID`:

```js
// Use of package.json for configuration
import { config, extensionManifest, version } from '../../package.json';

import { makeLog } from './utils';
import { selectors } from './store';
import { colorsWithAlpha } from './constants';

const log = makeLog('metrics');

const GA_URL = 'https://ssl.google-analytics.com/collect';
// eslint-disable-next-line prefer-destructuring
const GA_TRACKING_ID = config.GA_TRACKING_ID;
const GA_APP_ID = extensionManifest.applications.gecko.id;
const GA_APP_VERSION = version;
```

chuckharmston:
Starting work in on this, but I suspect I won't be finished by the time we have our standup this morning.

src/web/lib/storage.js
lmorchard: Doesn't really matter, as long as it's relatively unique

src/lib/metrics.js
lmorchard: Yeah, a polyfill might be a good idea

src/lib/metrics.js
lmorchard: I think that's just a check left over from some code I copied out of what had been used in a legacy add-on (and thus no `navigator` global). I can probably just remove the check on L317

.eslintrc.yaml
lmorchard: I can use destructuring like this, but I don't think it's especially better:
```javascript
const {
  version: GA_APP_VERSION,
  config: { GA_TRACKING_ID },
  extensionManifest: {
    applications: {
      gecko: {
        id: GA_APP_ID
      }
    }
  }
} = packageMeta;
```

lmorchard: Also just noticed on my screen the error's line number is in black, for some reason:

capture

But, since I have a line number now, I'm just going to punt on that lint rule for this chunk.

package.json
wresuolc: I'm waiting on a final name, and then will get a real GA ID and file an issue to implement it

lmorchard:
Added a polyfill for fetch() and some defensive coding around `navigator`

src/web/lib/components/AppHeader/index.js
```diff
@@ -16,23 +18,32 @@ export const AppHeader = ({ hasExtension, appVersion = null }) => {
         <div className="app-header__icon" />
         <header>
           <h1>Themer</h1>
-          <h3>A <a href="https://testpilot.firefox.com" target="_blank"
-           rel="noopener noreferrer">Firefox Test Pilot</a> Experiment</h3>
+          <h3>
```
pdehaan: nit: Not sure if the a11y linters called this out, but should this be an `h3` or an `h2`?

> _"Skipping heading ranks can be confusing and should be avoided where possible: Skipping heading ranks can be confusing and should be avoided where possible: Make sure that a `<h2>` is not followed directly by an `<h4>`, for example. It is ok to skip ranks when closing subsections, for instance, a `<h2>` beginning a new section, can follow an `<h4>` as it closes the previous section."_
>
> &mdash; via https://www.w3.org/WAI/tutorials/page-structure/headings/

package.json
chuckharmston: > Sadly it's in package.json so we can't add a `// TODO` comment.

And @linuxwolf has worked very hard to keep it that way 😉

src/web/lib/components/AppHeader/index.js
lmorchard: Not really a part of this PR or issue... that just got reformatted with pretty

src/lib/metrics.js
```diff
@@ -0,0 +1,340 @@
...
+// becomes available
+const INSTALL_FAILURE_DELAY = 10 * 1000;
+let installTimer = null;
+
+// See also: https://github.com/mozilla/ThemesRFun/blob/master/docs/metrics.md
+
+// Custom metrics
+let cm1; // integer count of select-full events from start of visit
+let cm2; // integer count of select-bg events from start of visit
+let cm3; // integer count of select-color events from start of visit
```
chuckharmston: These aren't going to be the most useful of custom metrics, since looking for a sum is meaningless (a user who got to `cm3=3` will have a `cm3` sum of `6`), and thus means are also not...meaningful.

src/lib/metrics.js
```diff
@@ -0,0 +1,340 @@
+// Use of package.json for configuration
+import packageMeta from '../../package.json';
+
+import { makeLog } from './utils';
+import { selectors } from './store';
+import { colorsWithAlpha } from './constants';
+
+const log = makeLog('metrics');
+
+const GA_URL = 'https://ssl.google-analytics.com/collect';
```
chuckharmston: Any reason not to use [`testpilot-ga`](https://github.com/mozilla/testpilot-ga) here? It should be usable from websites (and is [in use by Send](https://github.com/mozilla/send/blob/f7dc86ab2b85aa745cd389b9f1a1b25d856f144e/app/metrics.js#L1)).

src/lib/metrics.js
lmorchard: I think that's more a concern for #11 and metrics.md - this PR is just implementing what we have so far, which we can iterate on if we revise the schema

src/lib/metrics.js
lmorchard: Mainly because I haven't seen testpilot-ga until just now, I'll take a look

lmorchard:
Switched to testpilot-ga and pulled out the redundant client UUID code

package.json
wresuolc: https://github.com/mozilla/Themer/issues/142 for the final ID

Issue #45, opened 2018-02-06
Ensure we have decent mobile styles


lmorchard:
From [acceptance criteria](https://github.com/mozilla/ThemesRFun/blob/master/docs/acceptance.md):

> - [ ] It should have a minimal, pleasant mobile experience that lets users know it's a desktop experiment.

youwenliang:
Here's an idea for the mobile page: (Need to think about the content)
https://mozilla.github.io/testpilot-assets/ThemesRFun/01_Measurements/Mobile_page/
Icon: https://github.com/mozilla/testpilot-assets/tree/master/ThemesRFun/01_Measurements/Mobile_page/assets

We could also add a call to action button that links to testpilot's experiment detail page.

johngruen:
This looks good to me, i can pick this up.

pdehaan:
@johngruen Is this still legit, or did you fix it all w/ PR #188?

johngruen:
No we should still do something for mobile browsers

PR #105, opened 2018-02-28
Switch banner to a Firefox download CTA in non-Firefox browsers

### LINKED ISSUE #28: Provide Firefox Download link if user is in a different browser

lmorchard:
Fixes #28

### DIFF
--- a/src/web/lib/components/App/index.js
+++ b/src/web/lib/components/App/index.js
@@ -45,6 +45,8 @@ const mapDispatchToProps = dispatch => ({
 });
 
 export const AppComponent = ({
+  isFirefox,
+  isMobile,
   addonUrl,
   urlEncodeTheme,
   clipboard,
@@ -74,7 +76,7 @@ export const AppComponent = ({
         <SharedThemeDialog {...{ pendingTheme, setTheme, clearPendingTheme }} />
       )}
     <AppBackground {...{ theme }} />
-    {!hasExtension && <Banner {...{ addonUrl, bottom: false }} />}
+    {!hasExtension && !isMobile && <Banner {...{ isFirefox, addonUrl, bottom: false }} />}
     {loaderDelayExpired && (
       <div className="app__content">
         <AppHeader {...{ hasExtension }} />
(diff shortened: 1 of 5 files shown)


src/web/lib/components/App/index.js
```diff
@@ -26,6 +26,7 @@ const mapStateToProps = state => {
     {}
   );
   return {
+    isFirefox: navigator.userAgent.toLowerCase().includes('firefox/'),
```
chuckharmston: Do we want to differentiate Fennec and iOS here?

src/web/lib/components/Banner/index.js
```diff
@@ -1,17 +1,29 @@
...
-    <div className="banner__content">
-      <span>Put the 🔥 in Firefox</span>
-      <a href={addonUrl} className="banner__button">
-        Install Themer
-      </a>
-    </div>
-    {!bottom && <div className="banner__spacer"/>}
-  </div>;
+export const Banner = ({ isFirefox, addonUrl, bottom = false }) => (
+  <div className={classNames('banner', { 'banner--bottom': bottom })}>
```
chuckharmston: We have `banner--bottom` and `banner__content` here. Could we standardize on a delimiter?

src/web/lib/components/App/index.js
lmorchard: Hmm, probably - though now that you mention it, I don't think we want to just display "Download firefox" for those browsers. Maybe we need a third message that pushes folks to desktop Firefox in the mobile case.

src/web/lib/components/Banner/index.js
lmorchard: Sure? Probably a @johngruen question, since this comes from an earlier commit and I just reformatted here

lmorchard:
Borrowed some user agent stuff from Test Pilot, hiding the banner altogether for mobile. We can circle back with a more specific CTA if necessary - maybe as part of #45

PR #107, opened 2018-02-28
67 eslint mozilla

lmorchard:
Security checklist encourages using the Mozilla eslint rules, so here we go. This looks like a scary-huge PR, but I may just self merge to tear off the band-aid. 

Broken into two commits:

1. [Switch over to Mozilla eslint ruleset](https://github.com/mozilla/ThemesRFun/pull/107/commits/8f639433f8448bf91d81f07fb3e18170a02340d4) - but with the double-quote rule disabled. Only one change to source is actually needed here, [to remove an optional parameter in an addEventListener call](https://github.com/mozilla/ThemesRFun/pull/107/commits/8f639433f8448bf91d81f07fb3e18170a02340d4#diff-46d3bd6e450a71561ea0f0ca69af5992L86).
1. [Turn the double-quote rule on](https://github.com/mozilla/ThemesRFun/pull/107/commits/1c3175f5f02d333a815cc5caf6881547004132aa) along with `npm run lint -- --fix` to convert all the single-quotes to double-quotes. No other manual changes.


.eslintrc.yaml
```diff
@@ -28,3 +28,20 @@ rules:
   no-use-before-define: off
   object-curly-newline: off
   react/prop-types: off
+
+overrides:
+  -
+    files: ['src/**/*.js']
+    globals:
+      process: true
+      require: true
```
pdehaan: Should we just set `env` to include `node` here (same as L46-47 below)?

lmorchard: Well, they aren't actually in a node environment and only needed those two globals to pass linting, so I figured I'd be conservative. The webpack config files on L46-47 *are* node modules

.eslintrc.yaml
```diff
@@ -1,13 +1,13 @@
+extends:
+- eslint:recommended
+- plugin:import/recommended
+- plugin:mozilla/recommended
+- plugin:react/recommended
+
 env:
   browser: true
   es6: true
   jquery: true
```
pdehaan: Do we use jquery? Or was this leftover from earlier builds or airbnb-base?

lmorchard:
> I'm not going to look at 1c3175f, and you can't make me.

😨 there be <strike>dragons</strike> quotes

PR #111, opened 2018-03-02
Move default and preset themes into a directory of JSON files

lmorchard:


src/lib/themes.js
```diff
@@ -0,0 +1,12 @@
...
+  false,
+  /.*\.json/
+);
+
+export const defaultTheme = presetThemesContext("./default.json");
+
+export const presetThemes = presetThemesContext
+  .keys()
+  .map((filename, idx) => ({ idx, filename, ...presetThemesContext(filename) }))
+  .sort(({ filename: a }, { filename: b }) => a.localeCompare(b));
```
lmorchard: Just sorting by filename right now, but we could change this to whatever order

src/preset-themes/001.json
```diff
@@ -0,0 +1,35 @@
+{
+  "title": "001",
```
lmorchard: Threw in a title for the heck of it. Filenames & titles are 🤷‍♂️ because I wasn't feeling creative. We can tweak.

src/web/lib/components/PresetThemeSelector/index.js
```diff
@@ -2,21 +2,21 @@ import React from "react";
...
     <h2>Choose a preset theme</h2>
-    {presetColors.map((colors, themeId) => {
+    {presetThemes.map((colors, themeId) => {
       const theme = {
         images: "",
         colors
       };
       return (
-        <div key={themeId} className="preset-theme-preview">
+        <div key={themeId} title={colors.title} className="preset-theme-preview">
```
lmorchard: Punted on adding the title to the UI in any way other than as a hover-title

pdehaan:
> also didn't add any linting or validation for JSON

I ran this through my local Joi linter from the other PR and everything worked fine [after I added `title` to the schema].

Only nit I had was the "Default" theme is currently the 5th theme in the list, and it'd move every time we add more presets. Not sure if we should name the file "./src/preset-themes/000.json" in hopes that it'd always be the first theme in the list.

lmorchard:
> Only nit I had was the "Default" theme is currently the 5th theme in the list, and it'd move every time we add more presets.

Yeah, that's one of the things I punted on in case we want to do more work-shopping of the names. We could even just pick a theme at random on every cold load if we wanted to

src/web/lib/components/PresetThemeSelector/index.js
pdehaan: LGTM. I tried hacking a title to add an `onmouseover=""` handler in the theme JSON, but looks like React automatically escapes that XSS nonsense:

```
"title": "Panini\" onmouseover=\"alert(1)",
```

PR #129, opened 2018-03-08
Add logger

johngruen:
I did this on top of my pagination tweaks branch, so maybe we should merge that first.

Added a little logger component that will show up on dev to make getting themes out a bit easier.

### DIFF
--- a/src/web/lib/_common.scss
+++ b/src/web/lib/_common.scss
@@ -11,7 +11,7 @@ $grid: 4px;
   justify-content: center;
   text-decoration: none;
   transition: background 50ms;
-  min-width: 180px;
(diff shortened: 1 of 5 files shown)


src/web/index.js
```diff
@@ -144,7 +144,8 @@ render(
       clipboard,
       storage,
       isMobile,
-      isFirefox
+      isFirefox,
+      debug: DEBUG
```
lmorchard: Don't really need to pass this as a prop, IMO - just import DEBUG in the components and use it directly. It's a build-time thing, which webpack can actually exclude from the bundled JS with tree-shaking if DEBUG is false. Not a hugebig deal though

PR #130, opened 2018-03-08
pixel snap assets

johngruen:
Notice the icons were not pixel snapped


johngruen:
@chuckharmston They were! just force pushed a fix to #132

johngruen:
@chuckharmston 🤷‍♂️

PR #141, opened 2018-03-08
Switch use of headerURL to additional_backgrounds in themes

### LINKED ISSUE #126: Reconsider the use of headerURL in the web app/preset themes
So, in the app, we're using `headerURL` to refer to images, but in the add-on we're actually displaying images through the `additional_backgrounds` array. In order to better align the vocabulary we're using in our preset themes, the site, and the add-on, we should change preset themes so that the images section has no headerURL, but does optionally have an additional_backgrounds array. Also, we go through the app components and scrub any references to the headerURL in favor of `additional_backgrounds`

lmorchard:
Also with a few tweaks to make saved theme handling more robust & gracefully convert existing data to new structure.

Fixes #126


src/extension/background.js
```diff
@@ -50,10 +51,13 @@ const storeTheme = ({ theme }) => browser.storage.local.set({ theme });
...
 const applyTheme = ({ theme }) => {
   log("applyTheme", theme);
-  if (!theme) { return; }
+  if (!theme) {
+    return;
+  }
 
-  const backgroundImage = bgImages.keys().includes(theme.images.headerURL)
-    ? bgImages(theme.images.headerURL)
+  const background = theme.images.additional_backgrounds[0];
```
pdehaan: Do we need to check if `theme.images.additional_backgrounds` isn't `undefined` or `null` here before trying to access the 0 index? (we seem to below in [./src/lib/metrics.js:116-118](https://github.com/mozilla/Themer/pull/141/files#diff-d0154faa85aed1ddcb0a1d33f0484caeR116))

lmorchard: Should be okay - there's a call to [normalizeTheme](https://github.com/lmorchard/Themer/blob/311ede9fbfb3ed66fe4465ac25d1cd0148c3ef47/src/extension/background.js#L35) on the way in from the web, and the default is bg-000.svg if the set of dynamic imports don't contain the value

johngruen:
looks great!

PR #144, opened 2018-03-09
add some patterns

johngruen:


src/preset-themes/005.json
```diff
@@ -0,0 +1,43 @@
...
+      "h": 32,
+      "s": 0,
+      "l": 99,
+      "a": 36
+    },
+    "toolbar_field_text": {
+      "h": 38,
+      "s": 96,
+      "l": 65,
+      "a": 100
```
pdehaan: nit: only `toolbar` and `toolbar_field` support alpha channels, per:
https://github.com/mozilla/Themer/blob/cf279fac5a7f1a81a937884a8569c61215585c42/src/lib/constants.js#L16-L19

src/web/lib/components/ThemeBackgroundPicker/index.js
```diff
@@ -57,6 +57,7 @@ class ThemeBackgroundPicker extends React.Component {
     const backgroundSwatch = theme.images.headerURL
       ? `url(${bgImages(theme.images.headerURL)})`
       : "";
+      console.log(backgroundSwatch); //eslint-disable-line
```
pdehaan: nit: feels debug-y.

src/web/lib/components/ThemeBackgroundPicker/index.js
```diff
@@ -7,7 +7,7 @@ import Metrics from "../../../../lib/metrics";
 
 import "./index.scss";
 
-const bgImages = require.context("../../../../images/", false, /bg-.*\.png/);
+const bgImages = require.context("../../../../images/patterns/", false, /bg-.*\.svg/);
```
pdehaan: I don't know enough about the internal magic of Webpack's `require.context()`, or if it's possible to save the /images/patterns/*.svg array in /src/lib/constants.js since it looks like we're repeating this line in 3 places in this PR.

¯\\\_(ツ)\_/¯

src/images/patterns/bg-0.svg
```diff
@@ -0,0 +1 @@
+<svg width="1" height="1" viewBox="0 0 1 1" xmlns="http://www.w3.org/2000/svg"><title>bg-0</title></svg>
```
pdehaan: Not sure if we want to potentially rename these to `bg-00.svg` since it seems like the ordering looks like:

1. bg-0.svg
1. bg-10.svg
1. bg-11.svg
1. ...
1. bg-19.svg
1. bg-2.svg
1. bg-20.svg

Currently as we add more themes the order will always slightly shuffle, where new themes are added in the middle.

src/preset-themes/005.json
```diff
@@ -0,0 +1,43 @@
...
+    }
+  },
+  "images": {
+    "headerURL": "./bg-16.svg"
+  }
+}
```
pdehaan: Also, not sure if it's strictly "required" or not, but this file is missing a `title` attribute (says my 🗑:fire: schema linter).

src/images/patterns/bg-0.svg
johngruen: yeah good call

src/web/lib/components/ThemeBackgroundPicker/index.js
johngruen: this sounds like a @lmorchard question since webpack still seems mostly terrifyingly opaque to me...maybe file a separate issue?

johngruen:
huh, so i previously had bg-36.svg as my background (changed to bg-036.svg) in my latest force push to add a little more left-padding to the pattern file names. This caused the whole thing to error. Maybe I'll change back to two decimal place names for now

pdehaan:
Well, I think we'd only need "bg-_nnn_.svg" if you were planning on having more than 100 patterns, which seems somewhat of a high and unmanageable number. I was only thinking of going "bg-_nn_.svg" and saying we limit it to 99 patterns.

But now I'm fascinated at what your error was. Only thing I can immediately think of is that your theme metadata had specified the active background was bg-36.svg and renaming the file to bg-036.svg would cause it to go `undefined` or 404 or some other bizarre thing.

lmorchard:
> This caused the whole thing to error. Maybe I'll change back to two decimal place names for now

What was the error?

johngruen:
so i had a bg set the filename of which no longer existed

src/web/lib/components/ThemeBackgroundPicker/index.js
lmorchard: Actually, what might be a good idea is to move `bgImages` into `src/lib/themes.js` as a shared export since we use it both in the add-on and the website. But, I wouldn't block this PR to do it.

johngruen:
hang on a sec and I'll repro and drop the log in

lmorchard:
> so i had a bg set the filename of which no longer existed

Not a thing for this PR - but if we have an error on an invalid bg image filename, that's an issue worth fixing. Want to be able to handle whatever the ?theme param and saved themes throw at us without error

src/images/patterns/bg-000.svg
```diff
@@ -0,0 +1 @@
+<svg width="1" height="1" viewBox="0 0 1 1" xmlns="http://www.w3.org/2000/svg"><title>bg-000</title></svg>
```
pdehaan: One last comment, because I am a garbage human being...
I think we'll also need to address this, since we're in the spirit of file renaming:
https://github.com/mozilla/Themer/blob/cf279fac5a7f1a81a937884a8569c61215585c42/src/extension/background.js#L57

Not sure if @lmorchard or anybody else knows of any other hard-coded file references or defaults. This was the only obvious thing that I could find doing some <kbd>$ git grep "bg-" | grep ".js:"</kbd>.

johngruen:
err, i was in meetings, lemme reexport the files real quick...

example of what a missing file does:
```
Error: Cannot find module './bg-045.svg'.
.svg:54
The above error occurred in the <ThemeBackgroundPicker> component:
    in ThemeBackgroundPicker (created by OnClickOutside(ThemeBackgroundPicker))
    in OnClickOutside(ThemeBackgroundPicker) (created by AppComponent)
    in div (created by AppComponent)
    in section (created by BrowserPreview)
    in div (created by BrowserPreview)
    in BrowserPreview (created by AppComponent)
    in main (created by AppComponent)
    in div (created by AppComponent)
    in AppComponent (created by Connect(AppComponent))
    in Connect(AppComponent)
    in Provider

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
react-dom.development.js:9747
Error: Cannot find module './bg-045.svg'.
.svg:54
```

pdehaan:
> Not a thing for this PR - but if we have an error on an invalid bg image filename, that's an issue worth fixing. Want to be able to handle whatever the ?theme param and saved themes throw at us without error

Not sure if that'd be related to https://github.com/mozilla/Themer/issues/145#issuecomment-371750443 in any way (a few promises w/o a `.catch()`, so we may leak a few errors). Also, my local Joi schema linter does a hacky check to see if the specified .png/.svg is actually a real file in ./src/images/patterns/*, but that may not save us if people are fiddling w/ ?theme params.

lmorchard:
Created #147 to catch the bg image error thing as a task

PR #149, opened 2018-03-09
Use addons-linter in CircleCI linting to catch AMO rejections

lmorchard:
This PR should fail tests, because the add-on linting should pick up logo.svg not being square. Will fix with a rebase on master assuming that gets fixed soon


lmorchard:
One more tweak (which should also fail) and then I'll rebase against master (which should pass)

Issue #79, opened 2018-02-23
Add twitter/opengraph headers?


pdehaan:
Not sure if these came up in the a11y linting bug results from **sonarwhal** or **lighthouse** tools in https://github.com/mozilla/ThemesRFun/issues/47#issuecomment-368087825 (or anywhere else), but...

Do we want to add [Twitter](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup) and [OpenGraph](http://ogp.me/) meta tags for improved sharing greatness?

```html
<head>
  <meta charset="utf-8">
  <title>Firefox Theme Generator</title>
  <link rel="icon" type="image/svg+xml" href="images/icon.svg">
  <link href="index.css" rel="stylesheet">
  <script type="text/javascript" charset="utf-8" async="" src="https://mozilla.github.io/ThemesRFun/vendor/json-url-msgpack.js"></script>
  <script type="text/javascript" charset="utf-8" async="" src="https://mozilla.github.io/ThemesRFun/vendor/json-url-lzma.js"></script>
  <script type="text/javascript" charset="utf-8" async="" src="https://mozilla.github.io/ThemesRFun/vendor/json-url-safe64.js"></script>
</head>
```

Shruti9520:
@pdehaan @lmorchard Can I work on this? :-)

PR #157, opened 2018-03-10
Fixes-79-Add twitter/opengraph meta tags

Shruti9520:
Fixes [#79](https://github.com/mozilla/Themer/issues/79)
File changed in this PR-
1. `src/web/index.html.ejs`


lmorchard:
I think this looks pretty good, thanks!

PR #160, opened 2018-03-11
Add check to see if Themer tab already exists before creating new one

### LINKED ISSUE #119: [Enhancement] The "Firefox Theme Generator" tab should be focused if already open when the "ThemesRFun" toolbar button is pressed
**[Affected versions]:**
- Firefox 58.0.2 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "ThemesRFun" add-on (0.0.24 custom build from 07.03.2018) installed.

**[Steps to reproduce]:**
2. Press the "ThemesRFun" toolbar button and open/focus another tab.
3. Press the "ThemesRFun" toolbar button again and observe the behavior.

**[Expected result]:**
- The "Firefox Theme Generator" tab is focused.

**[Actual result]:**
- A new "Firefox Theme Generator" tab is opened.

**[Notes]**:
- If an "about:addons"/"about:preferences" tab is already opened, clicking the "Add-ons"/"Options" button will focus the existing tab. Considering this we think that we should do the same with the "Firefox Theme Generator" tab.
rec of the issue

cedricium:
Fixes #119. Currently, when the browserAction button is pressed, a new Themer generator tab is always created. As an enhancement, these changes check the current window's tabs for a Themer tab and if found, sets the Themer tab to be active / focused. If a Themer tab is not found, then one is created like the original implementation.

In order to query the window's tabs, the `tabs` permission had to be added.

**Demo:**
themer_119

### DIFF
--- a/src/extension/background.js
+++ b/src/extension/background.js
@@ -12,9 +12,18 @@ const bgImages = require.context("../images/patterns/", false, /bg-.*\.svg/);
 const siteUrl = process.env.SITE_URL;
 
 const init = () => {
-  browser.browserAction.onClicked.addListener(() =>
-    browser.tabs.create({ url: siteUrl })
-  );
+  browser.browserAction.onClicked.addListener(() => {
+    browser.tabs.query({ currentWindow: true })
+      .then(tabs => {
+        const themerTab = tabs.find(tab => {
+          return tab.title === "Firefox Themer";
+        });
+        if (themerTab)
+          browser.tabs.update(themerTab.id, { active: true });
+        else
+          browser.tabs.create({ url: siteUrl });
+      });
+  });
   browser.runtime.onConnect.addListener(port => {
     port.onMessage.addListener(messageListener(port));
     port.postMessage({ type: "hello" });
--- a/src/extension/background.js
+++ b/src/extension/background.js
@@ -15,13 +15,12 @@ const init = () => {
   browser.browserAction.onClicked.addListener(() => {
     browser.tabs.query({ currentWindow: true })
       .then(tabs => {
-        const themerTab = tabs.find(tab => {
-          return tab.title === "Firefox Themer";
-        });
-        if (themerTab)
+        const themerTab = tabs.find(tab => tab.url.includes(siteUrl));
+        if (themerTab) {
           browser.tabs.update(themerTab.id, { active: true });
-        else
+        } else {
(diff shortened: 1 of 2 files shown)


src/extension/background.js
```diff
@@ -12,9 +12,18 @@ const bgImages = require.context("../images/patterns/", false, /bg-.*\.svg/);
...
 
 const init = () => {
-  browser.browserAction.onClicked.addListener(() =>
-    browser.tabs.create({ url: siteUrl })
-  );
+  browser.browserAction.onClicked.addListener(() => {
+    browser.tabs.query({ currentWindow: true })
+      .then(tabs => {
+        const themerTab = tabs.find(tab => {
+          return tab.title === "Firefox Themer";
```
lmorchard: Two suggestions here:
* I'd rather we searched on `tab.url === siteUrl` here. Someday we may translate this title into other languages and so this string comparison will no longer work
* This could be made more compact like so: `const themerTab = tabs.find(tab => tab.url === siteUrl);`

src/extension/background.js
```diff
@@ -12,9 +12,18 @@ const bgImages = require.context("../images/patterns/", false, /bg-.*\.svg/);
...
+        });
+        if (themerTab)
```
lmorchard: We'd prefer braces be used at all times on conditionals - i.e. `if (themerTab) { ... } else { ... }`

src/extension/background.js
cedricium: @lmorchard I agree, comparing against the url would be better. Since `siteUrl` is just the base url, would it be better to do:
```js
const themerTab = tabs.find(tab => tab.url.includes(siteUrl));
```
since there could be a `?theme` query in the url?

src/extension/background.js
pdehaan: I think that's the [`curly`](https://eslint.org/docs/rules/curly) rule in ESLint.

src/extension/background.js
lmorchard: Oh, good catch! Yeah, `includes` is better than `=== ` - using `.indexOf(siteUrl) === 0` would be even more exact, though it's unlikely we'll find the siteUrl elsewhere in the url for a false positive

lmorchard:
Gave this another run through, looks good to me now!

PR #161, opened 2018-03-12
Enable use of PORT env var to change port in both site and extension

lmorchard:


lmorchard:
Rebased against master and added a blurb to README.md about the env var

README.md
```diff
@@ -28,6 +28,16 @@
...
+**Note:** If you have problems seeing the editor at `http://localhost:8080/` on
+your computer, it's possible that you already have some other service using
+port 8080. You can change the port that Themer uses for local development:
+
+* For Linux & OS X: `PORT=9090 npm start`
+* For Windows: `.\node_modules\.bin\cross-env PORT=9090 npm start`
+
+This example switches to port 9090, but you can supply a different port as
+needed.
+
```
chuckharmston: Wouldn't be sad to see `SITE_URL` also  documented here.

lmorchard: It is actually mentioned down here for deployment:

https://github.com/mozilla/Themer/blob/master/README.md#build--release

PR #163, opened 2018-03-12
Start using greenkeeper-lockfile on CircleCI

lmorchard:
May self merge this if it looks like it works on CircleCI. Winging it slightly because they only document CircleCI 2.0, and this project hasn't moved there yet


lmorchard:
Seems to pass on a normal PR - the proof will be what this does on the next Greenkeeper PR

Issue #108, opened 2018-03-01
Add a theme linter?


pdehaan:
Not sure if we want to build something w/ [**ajv**](http://npm.im/ajv), or [**joi**](http://npm.im/joi), or your schema validator of choice to make sure we're validating themes so that they don't have any extra/missing properties. Will also make it easier to validate theme data in the future if we add names (#91) or additional properties (#70).

I had a couple lame ideas in https://github.com/mozilla/ThemesRFun/pull/106#issuecomment-369137906.

lmorchard:
This could be a good idea to go along with separating out the themes as individual JSON files from constants.js

pdehaan:
I also found it a little odd that the `defaultColors` constant has the exact same values, as `presetColors[0]`.

https://github.com/mozilla/ThemesRFun/blob/a57c5b5a231e669193662091138ea5bd8c963e27/src/lib/constants.js#L21-L54

https://github.com/mozilla/ThemesRFun/blob/a57c5b5a231e669193662091138ea5bd8c963e27/src/lib/constants.js#L56-L90

Not sure if it'd be better to just do:

```js
export const presetColors = [
  {...defaultColors},
  blah,
};
```

I was trying to use [**glob**](http://npm.im/glob) in ./src/lib/constants.js to grab my extracted preset themes from ./src/lib/presets/\*.json, but it was seemingly failing somewhere since I don't think I'm running in a Node context in ./src/lib/*. Using **glob** otherwise worked great when running my schema linter locally via `$ babel-node schema-lint-joi`. Although explicitly importing each preset manually using something like this worked in ./src/lib/constants.js:

```js
import preset1 from './presets/1.json';
import preset2 from './presets/2.json';
import preset3 from './presets/3.json';
import preset4 from './presets/4.json';
import preset5 from './presets/5.json'; // Hotdog Stand 🌭
```

lmorchard:
> I also found it a little odd that the defaultColors constant has the exact same values, as presetColors[0].

`defaultColors` predates the preset colors feature just slightly, so I guess it was just a good copypasta candidate.

> I was trying to use glob in ./src/lib/constants.js to grab my extracted preset themes from ./src/lib/presets/*.json, but it was seemingly failing somewhere since I don't think I'm running in a Node context in ./src/lib/*. 

[...]

PR #172, opened 2018-03-14
Fixes-108-Add a theme linter

Shruti9520:
Fixes [#108](https://github.com/mozilla/Themer/issues/108)
Files changed in this PR-
1.  `.eslintrc.yaml`
2. `package.json`
3. `src/preset-themes/005.json`

Files added in this PR-
1. ` bin/validate-themes.js`
2.  docs/theme-schema.json

### DIFF
--- a/bin/validate-themes.js
+++ b/bin/validate-themes.js
@@ -1,0 +1,28 @@
+const fs = require("fs");
+const path = require("path");
+const Ajv = require("ajv");
+
+const themesPath = path.join(__dirname, "..", "src", "preset-themes");
+const schemaFilename = path.join(__dirname, "..", "docs", "theme-schema.json");
+const schema = JSON.parse(fs.readFileSync(schemaFilename, "utf8"));
+
+const ajv = new Ajv({ allErrors: true });
+const validate = ajv.compile(schema);
+
+let foundInvalid = false;
+
+fs.readdirSync(themesPath)
+  .filter(filename => path.extname(filename) === ".json")
+  .forEach(filename => {
+    const data = fs.readFileSync(path.join(themesPath, filename), "utf8");
+    const theme = JSON.parse(data);
+    if (!validate(theme)) {
+      console.log("Theme validation failed for", filename, ajv.errorsText(validate.errors)); // eslint-disable-line no-console
+      foundInvalid = true;
+    }
+  });
+
+if (foundInvalid) {
+  console.log("Found invalid theme JSON data"); // eslint-disable-line no-console
+  process.exit(1);
+}
--- a/docs/theme-schema.json
+++ b/docs/theme-schema.json
@@ -1,0 +1,169 @@
+{
+    "properties": {
+        "title": {
+            "type": "string"
+        },
+        "colors": {
+            "type": "object",
+            "properties": {
+                "toolbar": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "integer",
+                            "maximum": 360,
+                            "minimum": 0
+                        },
+                        "s": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "l": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "a": {
+                            "type": "number",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l", "a"],
+                    "additionalProperties": false
+                },
+                "toolbar_text": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "integer",
+                            "maximum": 360,
+                            "minimum": 0
+                        },
+                        "s": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "l": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l"],
+                    "additionalProperties": false
+                },
+                "accentcolor": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "integer",
+                            "maximum": 360,
+                            "minimum": 0
+                        },
+                        "s": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "l": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l"],
+                    "additionalProperties": false
+                },
+                "textcolor": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "integer",
+                            "maximum": 360,
+                            "minimum": 0
+                        },
+                        "s": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "l": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l"],
+                    "additionalProperties": false
+                },
+                "toolbar_field": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "integer",
+                            "maximum": 360,
+                            "minimum": 0
+                        },
+                        "s": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "l": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "a": {
+                            "type": "number",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l", "a"],
+                    "additionalProperties": false
+                },
+                "toolbar_field_text": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "integer",
+                            "maximum": 360,
+                            "minimum": 0
+                        },
+                        "s": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "l": {
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l"],
+                    "additionalProperties": false
+                }
+            },
+            "required": ["toolbar", "toolbar_text", "accentcolor", "textcolor", "toolbar_field", "toolbar_field_text"],
+            "additionalProperties": false
+        },
+        "images": {
+            "type": "object",
+            "properties": {
+                "headerURL": {
+                    "type": "string"
+                },
+                "additional_backgrounds": {
+                    "type": "array"  
+                }
+            }
+        }
+    },
+    "required": ["title", "colors"],
+    "additionalProperties": false
+}
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -40,7 +40,7 @@ overrides:
     env:
       webextensions: true
   -
-    files: ['webpack.*.js']
+    files: ['webpack.*.js','bin/*.js']
     rules:
       import/unambiguous: off
     env:
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "sign": "npm-run-all clean:extension build:extension && web-ext sign --source-dir=build/extension && mv web-ext-artifacts/*.xpi ./addon.xpi",
     "run": "npm run build:extension && web-ext run --source-dir=build/extension",
     "lint": "npm-run-all lint:*",
+    "lint:themes": "node ./bin/validate-themes.js",
     "lint:css": "stylelint 'src/**/*.scss'",
     "lint:js": "eslint --color .",
     "validate": "npm-run-all validate:*",
--- a/src/preset-themes/005.json
+++ b/src/preset-themes/005.json
@@ -1,6 +1,6 @@
 {
+  "title": "005",
   "colors": {
-    "title": "005",
     "toolbar": {
       "h": 33,
       "s": 72,


Shruti9520:
@lmorchard @pdehaan 
I use the following code to test each of the file-
`require('fs').readFile('../001.json', 'utf8', function(err, data) {
    if (err) {
        console.log(err);
    } else {
        var obj1 = JSON.parse(data);
        test(obj1);
    }
});`
This is definately not a good approach. I would be great if you suggest me something better for this. :-)

Thanks!

src/preset-themes/jsonSchema/jsonSchema.js
```diff
@@ -0,0 +1,227 @@
...
+                        "l": {
+                            "type": "number",
+                            "maximum": 100,
+                            "minimum": 0
+                        },
+                        "a": {
+                            "type": "number",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
```
pdehaan: Only 2 properties support an alpha channel, so we should remove the `a` where it isn't needed (and possibly throw errors when it isn't wanted but is present):

https://github.com/mozilla/Themer/blob/38933c93a6c304a70b095dfd737b3b2c8ce871d6/src/lib/constants.js#L16-L19

Similarly, we should make sure that the `toolbar` and `toolbar_field` properties have the `"a"` property marked as required.

src/preset-themes/jsonSchema/jsonSchema.js
```diff
@@ -0,0 +1,227 @@
...
+var ajv = new Ajv({
+    allErrors: true
+});
+
+var schema = {
+    "properties": {
+        "title": {
+            "type": "string"
+        },
+        "colors": {
```
pdehaan: I didn't see this below in the schema, but we should mark "title" and "colors" as required properties.

chuckharmston:
A few more general notes:

1. We should separate the linter from the task that does the linting.
2. We should make the passing of linting be required for our CI (see [CircleCI 2.0 docs](https://circleci.com/docs/2.0/))
3. Since node will always be running this task, we can use [`fs`](https://nodejs.org/api/fs.html) to traverse the `src/preset-themes` directory to make sure that every theme is always validated.
4. The files introduced here should probably not live in `src/preset-themes`. @lmorchard might offer some suggestions on better places for them to live?

src/preset-themes/jsonSchema/jsonSchema.js
```diff
@@ -0,0 +1,227 @@
...
+            "type": "object",
+            "properties": {
+                "toolbar": {
+                    "type": "object",
+                    "properties": {
+                        "h": {
+                            "type": "number",
```
pdehaan: small nit: "h", "s", and "l" should be integers, and I think only "a" supports a floating point value, per:

https://github.com/mozilla/Themer/blob/38933c93a6c304a70b095dfd737b3b2c8ce871d6/src/preset-themes/001.json#L25-L30

src/preset-themes/jsonSchema/jsonSchema.js
```diff
@@ -0,0 +1,227 @@
...
+                }
+            },
+            "required": ["toolbar", "toolbar_text", "accentcolor", "textcolor", "toolbar_field", "toolbar_field_text"]
+        }
+    }
+};
+
+var validate = ajv.compile(schema);
+
+require("fs").readFile("../001.json", "utf8", function(err, data) {
```
pdehaan: We should do something like `const fs = require("fs");` at the top of the file, then you could modify this to `fs.readFile(...)`.

src/preset-themes/jsonSchema/jsonSchema.js
```diff
@@ -0,0 +1,227 @@
+import { makeLog } from "../../lib/utils";
+
+const log = makeLog("background");
```
pdehaan: @lmorchard Do you think we need `makeLog(...)`, or since this is more of a CLI script, should we just use `console.log()` to dump to stdout and `process.exitCode`?

src/preset-themes/jsonSchema/jsonSchema.js
```diff
@@ -0,0 +1,227 @@
...
+    }
+});
+require("fs").readFile("../005.json", "utf8", function(err, data) {
+    if (err) {
+        log(err);
+    } else {
+        var obj5 = JSON.parse(data);
+        test(obj5);
+    }
+});
```
pdehaan: Instead of `fs.readFile()` and parsing as JSON, one approach (not great, but it's _an_ approach) would be to require the JSON file using a relative path:

```js
const validate = ajv.compile(schema);

const theme001 = require("../001.json");
test(theme001);

const theme002 = require("../002.json");
test(theme002);

const theme003 = require("../003.json");
test(theme003);

const theme004 = require("../004.json");
test(theme004);

const theme005 = require("../005.json");
test(theme005);

function test(data) {
  const valid = validate(data);
  if (valid) {
    console.log("Valid!");
  } else {
    console.log("Invalid: " + ajv.errorsText(validate.errors));
  }
}
```

Still not great, since you'd have to declare each file specifically, and you're currently only linting 5 of the 7 themes in that directory.

Another possible option is to use `fs.readdir()` or `fs.readdirSync()` and do something like this:

```js
const validate = ajv.compile(schema);

fs.readdirSync("./src/preset-themes")
  .filter(file => file.endsWith(".json"))
  .forEach(file => test(file));

function test(file) {
  const data = require(`../${file}`);
  const valid = validate(data);
  if (valid) {
    console.log(`${file}: Valid!`);
  } else {
    console.log(`${file}: Invalid: ${ajv.errorsText(validate.errors)}`);
    process.exitCode = 1;
  }
}
```

... which gives me the following output:

```sh
➜  ThemesRFun git:(PR172) ✗ node ./src/preset-themes/jsonSchema/jsonSchema.js
001.json: Valid!
002.json: Valid!
003.json: Valid!
004.json: Valid!
005.json: Valid!
default.json: Valid!
hotdog.json: Valid!
```

Not sure if it works 100%, we'd have to test it with some theme errors, but may get you a bit closer.

Final note is that everything seems to lint as valid, so I think we're still missing some pieces, as theme 005 is still invalid (missing `title`, bonus `title` property in the wrong place, schema isn't complaining we accept `a` fields where we shouldn't, and some themes may have an `images` object which has a `headerURL`:

https://github.com/mozilla/Themer/blob/38933c93a6c304a70b095dfd737b3b2c8ce871d6/src/preset-themes/005.json#L37-L39

Or an `images` object which has an `additional_backgrounds` array with 0-1 items:

https://github.com/mozilla/Themer/blob/38933c93a6c304a70b095dfd737b3b2c8ce871d6/src/preset-themes/default.json#L37-L39

But overall I think this is pretty close!

pdehaan:
And just to copy over @lmorchard's https://github.com/mozilla/Themer/issues/108#issuecomment-372825534 (which I believe answers @chuckharmston's 2nd bullet point in https://github.com/mozilla/Themer/pull/172#issuecomment-373098325 above):

> _"For this, you could just use [node's `fs` module](https://nodejs.org/dist/latest-v8.x/docs/api/fs.html#fs_fs_readfilesync_path_options) and read the JSON files directly, using `JSON.parse()` to parse them before testing them. We could run it alongside [other npm script commands we use for linting & validating](https://github.com/mozilla/Themer/blob/master/package.json#L31)."_

src/preset-themes/jsonSchema/jsonSchema.js
Shruti9520: Will do it. :-) 
Thanks for reviewing.

src/preset-themes/jsonSchema/jsonSchema.js
Shruti9520: okay

Shruti9520:
@chuckharmston Thanks for reviewing.
This is what I understood ( Please correct me if I am wrong at any point)-
1. I will make sure that `jsonSchema.js` file is separated from `preset-themes` directory.
2. To make the passing of linting be required for CI, I think I should add something in Package.json. Something like this https://github.com/mozilla/Themer/blob/master/package.json#L31. Isn't it @pdehaan ? :-)
3. For making use of `fs` to traverse the src/preset-themes directory to make sure that every theme is always validated, I will follow @pdehaan comments.

Thanks!

src/preset-themes/jsonSchema/jsonSchema.js
Shruti9520: @pdehaan Thank you so much!
I haven't set `additionalProperties: false` to any scope. So we can add additional objects like `"images": { 
   "headerURL": "./bg-016.svg" 
 } `

Also, for theme 005, lint should be invalid because we are interchanging `title` (which is actually of type `string` ) and `colors` (which is actually of type `object`).  When I checked it online, it was giving error like `Invalid: data.title should be string, data.colors should be object`, which is quite evident. I will check it once again and let you know the results. :-)

Thanks!

lmorchard:
Catching up with comments so far, good progress! Here's what I'd advise for next steps:

1. Move the validator script to `bin/validate-themes.js` from `src/preset-themes/jsonSchema/jsonSchema.js`
1. Extract the JSON schema in the script from `var schema = {` and move it into a standalone JSON file - maybe `docs/theme-schema.json`
1. Add an entry to `"scripts"` in `package.json` - something like this: 
    `"lint:themes": "node ./bin/validate-themes.js"`
1. Then, `bin/validate-themes.js` can look something like this:
```javascript
#!/usr/bin/env node
const fs = require("fs");
const path = require("path");
const Ajv = require("ajv");

const themesPath = path.join(__dirname__, "..", "src", "preset-themes");
const schemaFilename = path.join(__dirname__, "..", "docs", "theme-schema.json");
const schema = JSON.parse(fs.readFileSync(schemaFilename, "utf8"));

const ajv = new Ajv({ allErrors: true });
const validate = ajv.compile(schema);

fs.readdirSync(themesPath)
  .filter(filename => path.extname(filename) === ".json")
  .forEach(filename => {
    const data = fs.readFileSync(path.join(themesPath, filename), "utf8");
    const theme = JSON.parse(data);
    if (!validate(theme)) {
      console.error("Theme validation failed for", filename, ajv.errorsText(validate.errors)));
      process.exit(1);
    }
  });
```
This code is just kind of a suggestion off the top of my head, though. I haven't actually tried running it. So, you might want to tinker with it a bit to make it actually work. :grin:

Shruti9520:
@lmorchard Thanks for reviewing :-)

Shruti9520:
Hi @pdehaan @lmorchard I am getting this error after running `npm test`. I tried many solutions from the internet but it doesn't help. Errors like `require, path` is not defined is also coming.

binerror

lmorchard:
I think that error is because the current version of node.js doesn't fully support ES6. We use that syntax in the add-on and web source code because Babel transforms it into an earlier form of JavaScript. In this case, running directly in node, I don't think you can use makeLog this way. I would just go ahead and use `console.log()` rather than makelog

Shruti9520:
@lmorchard Thanks for quick reply.  After removing makeLog, I am getting this error- 
errortheer

lmorchard:
Oh yeah, I forgot about eslint - those errors should be pretty easy to fix:

1. Edit `.eslintrc.yaml` and look for line 43 near the bottom. Change `files: ['webpack.*.js']` to `files: ['webpack.*.js', 'bin/*.js']` - this will let lint know that scripts in the `bin/` directory should use a different set of rules.

2. Wherever `console.log()` is used in your script, add this comment at the end of the line `// eslint-disable-line no-console` - this will tell eslint that we intended to use console.log there and to ignore it

pdehaan:
@lmorchard Since we're using the `overrides` anyways in .eslintrc.yml, do you just want to disable the `no-console` rule for the webpack* and bin/*.js files?

https://github.com/mozilla/Themer/blob/6397cea7767b8cd7aec963412949f31625ec0175/.eslintrc.yaml#L42-L47

lmorchard:
Actually yeah, that's probably a better idea. Add a `no-console: off` under the `unambiguous` line along with adding `bin/*.js` to the files list and not bother adding the comments to the script

Shruti9520:
@lmorchard Yes, it is working. Thank you so much. :-)

lmorchard:
@Shruti9520 Thanks for working through this stuff!

Shruti9520:
@lmorchard @pdehaan So, my result is like this. It seems working fine. :-)

validatepr

docs/theme-schema.json
```diff
@@ -0,0 +1,156 @@
...
+                            "type": "integer",
+                            "maximum": 100,
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["h", "s", "l"],
+                    "additionalProperties": false
+                }
+            },
+            "required": ["toolbar", "toolbar_text", "accentcolor", "textcolor", "toolbar_field", "toolbar_field_text"]
```
pdehaan: I think we need `"additionalProperties": false` here too, to make sure we arent injecting something unexpected.

docs/theme-schema.json
```diff
@@ -0,0 +1,156 @@
...
+        }
+    },
+    "required": ["title", "colors"]
```
pdehaan: Same as above, we should add `"additionalProperties": false` here too (which should cause a few more linter errors, if my guess is correct).

bin/validate-themes.js
```diff
@@ -0,0 +1,30 @@
...
+fs.readdirSync(themesPath)
+  .filter(filename => path.extname(filename) === ".json")
+  .forEach(filename => {
+    const data = fs.readFileSync(path.join(themesPath, filename), "utf8");
+    const theme = JSON.parse(data);
+    if (!validate(theme)) {
+      console.log("Theme validation failed for", filename, ajv.errorsText(validate.errors)); // eslint-disable-line no-console
+      foundInvalid = true;
+    } else {
+      console.log(filename, "validate"); // eslint-disable-line no-console
```
pdehaan: I _think_ we can probably lose the `else {}` here. Not sure we need stdout verification that there were zero errors.

docs/theme-schema.json
Shruti9520: In that case both `default.json` and `005.json` is invalidated.

pdehaan:
> My CIRCLECI test fails because of 005.json file (invalidate json schema), which is absolutely fine. :-)

I'm super happy that Circle-CI is failing on this! Mainly because we caught a valid bug that I should have caught earlier.

```sh
> node ./bin/validate-themes.js

001.json validate
002.json validate
003.json validate
004.json validate
Theme validation failed for 005.json data should have required property 'title'
default.json validate
hotdog.json validate
Found invalid theme JSON data
```

That said, feel free to update 005.json theme and fix/move the incorrectly nested `title` property so Circle is happy again.

docs/theme-schema.json
pdehaan: Yay, my guess is correct!
We'll need to update the schema to allow for those other optional properties (`images.additional_backgrounds[]` and `images.headerURL`, IIRC).

bin/validate-themes.js
Shruti9520: Yeah. That would be better. :-) 
Thanks!

Shruti9520: @pdehaan Updated my Pull Request.

Shruti9520:
@pdehaan What about the extra `"images"` property in `005.json` and `default.json`? After adding `"additionalProperties": false` it is also throwing error.

pdehaan:
Again, great work on this PR and getting everything linted and passing!

I'll let @lmorchard give the final 👀 and press the shiny green button.

Shruti9520:
@lmorchard Will you please review it again? :-)

Thank you very much!

lmorchard:
Looks good! Have a couple further tweaks in mind but will open a follow up PR

PR #180, opened 2018-03-15
Populate opengraph/twitter meta tags from package.json

pdehaan:
No doubt this will be useless when we add l10n, but until then... 🎉

### DIFF
--- a/src/web/index.html.ejs
+++ b/src/web/index.html.ejs
@@ -3,17 +3,17 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width" />
-    <meta name="description" content="Theming demo for Firefox Quantum and beyond." />
+    <meta name="description" content="<%= htmlWebpackPlugin.options.description %>" />
     <!-- Open Graph meta tags-->
-    <meta property="og:title" content="Firefox Themer" />
-    <meta property="og:url" content="https://mozilla.github.io/Themer/" />
+    <meta property="og:title" content="<%= htmlWebpackPlugin.options.title %>" />
+    <meta property="og:url" content="<%= htmlWebpackPlugin.options.homepage %>" />
     <meta property="og:type" content="website" />
     <!-- Twitter card meta tags-->
-    <meta name="twitter:card" content="themer" />
+    <meta name="twitter:card" content="summary" />
     <meta name="twitter:site" content="@FxTestPilot" />
-    <meta name="twitter:title" content="Grab your parachute and get access to experimental Firefox features." />
-    <meta name="twitter:description" content="Theming demo for Firefox Quantum and beyond." />
-    <title>Firefox Themer</title>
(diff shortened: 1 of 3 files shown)


src/web/index.html.ejs
```diff
@@ -3,17 +3,17 @@
...
+    <meta name="description" content="<%= htmlWebpackPlugin.options.description %>" />
     <!-- Open Graph meta tags-->
-    <meta property="og:title" content="Firefox Themer" />
-    <meta property="og:url" content="https://mozilla.github.io/Themer/" />
+    <meta property="og:title" content="<%= htmlWebpackPlugin.options.title %>" />
+    <meta property="og:url" content="<%= htmlWebpackPlugin.options.homepage %>" />
     <meta property="og:type" content="website" />
     <!-- Twitter card meta tags-->
-    <meta name="twitter:card" content="themer" />
+    <meta name="twitter:card" content="summary" />
```
pdehaan: Not sure "themer" is a valid card type here. I stole "summary" from the TestPilot site's `<meta>` tags.

src/web/index.html.ejs
```diff
@@ -3,17 +3,17 @@
...
     <meta name="twitter:site" content="@FxTestPilot" />
-    <meta name="twitter:title" content="Grab your parachute and get access to experimental Firefox features." />
-    <meta name="twitter:description" content="Theming demo for Firefox Quantum and beyond." />
-    <title>Firefox Themer</title>
+    <meta name="twitter:title" content="<%= htmlWebpackPlugin.options.title %>" />
```
pdehaan: I'll have to re-read the docs to try and understand the difference betwixt a `twitter:title` and `twitter:description`, but tweak will change the `twitter:title` to "Firefox Themer" (versus the current parachute reference, which may have been a TestPilot relic).

webpack.web.js
```diff
@@ -24,7 +25,10 @@ module.exports = merge(common.webpackConfig, {
     new HtmlWebpackPlugin({
       template: "./src/web/index.html.ejs",
       filename: "index.html",
-      chunks: ["index"]
+      chunks: ["index"],
+      title: pkg.title,
+      description: pkg.description,
+      homepage: pkg.homepage
```
pdehaan: Conversely, I could just shim all of package.json values into the options, but that felt a _wee_ bit heavy handed, so I just injected the 3 params I "cared" about.

pdehaan:
This is all I know about l10n. I should probably file a separate bug about l10n status so we can discuss there, and put it in a milestone (if we're doing it for v1 or waiting)...

https://github.com/mozilla/Themer/blob/620412840e461a77c6c89df858731558e407e4e0/src/web/lib/components/AppHeader/index.scss#L71

lmorchard:
No localization plans yet, wanted to get to feature complete since things were kind of in experimental flux and strings would have been anything but stable

PR #183, opened 2018-03-16
176: dismiss color picker panel on esc/multiple clicks

### LINKED ISSUE #176: The "Color Picker" panel is not dismissed when its button is clicked or the "Esc" key is pressed
**[Affected versions]:**
- Firefox 58.0.2 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Themer" add-on (0.0.31 custom build) installed from the "about:debugging" page.

**[Steps to reproduce]:**
2. Click the "Themer" toolbar button.
3. Click the "Toolbar Icons and Text" bubbled text.
4. Click the bubble text and observe the behavior.

**[Expected result]:**
- The "Color Picker" panel is dismissed.

**[Actual result]:**
- Nothing happens.

**[Notes]:**
color picker dimiss

shr7:
Fixes #176

### DIFF
--- a/src/web/lib/components/ThemeBackgroundPicker/index.js
+++ b/src/web/lib/components/ThemeBackgroundPicker/index.js
@@ -36,6 +36,7 @@ class ThemeBackgroundPicker extends React.Component {
     this.state = {
       selected: false
     };
+    this.handleKeyPress = this.handleKeyPress.bind(this);
   }
 
   handleClick() {
@@ -45,6 +46,21 @@ class ThemeBackgroundPicker extends React.Component {
   handleClickOutside() {
     this.setState({ selected: false });
   }
+
+  handleKeyPress(event) {
+    if (event.keyCode === 27) {
+      this.setState({ selected: false });
+    }
+  }
+
+  componentDidMount() {
+    document.addEventListener("keydown", this.handleKeyPress);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener("keydown", this.handleKeyPress);
+  }
+
 
   render() {
     const { theme, setBackground } = this.props;
--- a/src/web/lib/components/ThemeBackgroundPicker/index.js
+++ b/src/web/lib/components/ThemeBackgroundPicker/index.js
@@ -4,6 +4,7 @@ import onClickOutside from "react-onclickoutside";
 
 import { colorToCSS, bgImages } from "../../../../lib/themes";
 import Metrics from "../../../../lib/metrics";
+import { ESC } from "../../../../lib/constants";
 import "./index.scss";
 
 const Background = ({
@@ -48,7 +49,7 @@ class ThemeBackgroundPicker extends React.Component {
   }
 
   handleKeyPress(event) {
-    if (event.keyCode === 27) {
+    if (event.keyCode === ESC) {
       this.setState({ selected: false });
     }
   }
--- a/src/web/lib/components/ThemeColorsEditor/index.js
+++ b/src/web/lib/components/ThemeColorsEditor/index.js
@@ -11,20 +11,14 @@ import "./index.scss";
 class ThemeColorsEditor extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
-      selected: false,
-      colorName: null
-    };
     this.handleKeyPress = this.handleKeyPress.bind(this);
   }
 
   handleClick(name) {
-    const { setSelectedColor } = this.props;
+    const { selectedColor, setSelectedColor } = this.props;
     setSelectedColor({ name });
-    if (this.state.selected && this.state.colorName !== name) {
-      this.setState({selected: this.state.selected, colorName: name });
-    } else {
-      this.setState({ selected: !this.state.selected, colorName: name });
+    if (selectedColor === name) {
+      setSelectedColor({ name: null });
     }
   }
 
@@ -33,12 +27,14 @@ class ThemeColorsEditor extends React.Component {
     if (selectedColor !== null) {
       setSelectedColor({ name: null });
     }
-    this.setState({ selected: false, colorName: null });
   }
 
   handleKeyPress(event) {
+    const { selectedColor, setSelectedColor } = this.props;
     if (event.keyCode === ESC) {
-      this.setState({ selected: false, colorName: null });
+      if (selectedColor !== null) {
+        setSelectedColor({ name: null });
+      }
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/web/lib/components/ThemeBackgroundPicker/index.js
```diff
@@ -46,6 +47,21 @@ class ThemeBackgroundPicker extends React.Component {
     this.setState({ selected: false });
   }
 
+  handleKeyPress(event) {
+    if (event.keyCode === 27) {
```
pdehaan: It's probably worth creating a `const` at the top of the file for the key code so we can use a more descriptive variable name.

For example, `const ESC = 27;`, so this can turn into `if (event.keyCode === ESC)`, or whatever.

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -9,21 +9,58 @@ import Metrics from "../../../../lib/metrics";
...
   handleClickOutside() {
     const { selectedColor, setSelectedColor } = this.props;
     if (selectedColor !== null) {
       setSelectedColor({ name: null });
     }
+    this.setState({ selected: false, colorName: null });
+  }
+
+  handleKeyPress(event) {
+    if (event.keyCode === 27) {
```
pdehaan: Same comment as above. Or since we're using it in two places now, not sure if there is a better place for shared constants.

pdehaan:
Random GitHub fact: if your initial commit message/comment contains _"Fixes #{issue_num}"_ or _"Closes #{issue_num}"_ the associated GitHub issue will be automatically closed when the PR is merged.

For example, if you edit your comment above to say _"Fixes #176"_ then #176 will automatically close when this gets merged.

src/web/lib/components/ThemeColorsEditor/index.js
shr7: I guess I should put it in lib/constants.js then

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -2,28 +2,65 @@ import React from "react";
...
+import { colorLabels, colorsWithAlpha, ESC } from "../../../../lib/constants";
 import { colorToCSS } from "../../../../lib/themes";
 import Metrics from "../../../../lib/metrics";
 
 import "./index.scss";
 
 class ThemeColorsEditor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
lmorchard: Hmm, I don't think we need to add local state to this component. Basically `colorName` should be the same as `this.props.selectedColor` and `selected` should match `this.props.selectedColor !== null`. So we should be able to get the info needed just from the supplied props

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -2,28 +2,65 @@ import React from "react";
...
+      selected: false,
+      colorName: null
+    };
+    this.handleKeyPress = this.handleKeyPress.bind(this);
+  }
+
+  handleClick(name) {
+    const { setSelectedColor } = this.props;
+    setSelectedColor({ name });
+    if (this.state.selected && this.state.colorName !== name) {
```
lmorchard: Here, we should be able to check if `name === this.props.selectedColor` to decide whether to `setSelectedColor({ name: null })` which should clear the current selected color

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -2,28 +2,65 @@ import React from "react";
...
     const { selectedColor, setSelectedColor } = this.props;
     if (selectedColor !== null) {
       setSelectedColor({ name: null });
     }
+    this.setState({ selected: false, colorName: null });
+  }
+
+  handleKeyPress(event) {
+    if (event.keyCode === ESC) {
+      this.setState({ selected: false, colorName: null });
```
lmorchard: I think `setSelectedColor({ name: null })` on ESC would be simpler here - basically copying what `handleClickOutside` does without local state

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -36,8 +73,8 @@ class ThemeColorsEditor extends React.Component {
             return [
               <li
                 key={`dt-${idx}`}
-                className={classnames(name, "color", { selected: selectedColor === name })}
-                onClick={() => setSelectedColor({ name })}
+                className={classnames(name, "color", { selected: selected && selectedColor === name })}
+                onClick={this.handleClick.bind(this, name)}
```
lmorchard: Probably clearer to just use `onClick={() => this.handleClick(name)}` here, since we need to make a closure to pass along `name` every time anyway

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -2,26 +2,55 @@ import React from "react";
...
 
 class ThemeColorsEditor extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleKeyPress = this.handleKeyPress.bind(this);
+  }
+
+  handleClick(name) {
+    const { selectedColor, setSelectedColor } = this.props;
+    setSelectedColor({ name });
```
lmorchard: Could make the next 4 lines simpler with a single `setSelectedColor` call like this:
```javascript
setSelectedColor({
  name: selectedColor === name ? null : name 
});
```

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -2,26 +2,55 @@ import React from "react";
...
   handleClickOutside() {
     const { selectedColor, setSelectedColor } = this.props;
     if (selectedColor !== null) {
       setSelectedColor({ name: null });
     }
   }
 
+  handleKeyPress(event) {
+    const { selectedColor, setSelectedColor } = this.props;
+    if (event.keyCode === ESC) {
```
lmorchard: Might as well combine these two if statements like:
```javascript
if (event.keyCode === ESC && selectedColor !== null) {
```

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -36,8 +65,8 @@ class ThemeColorsEditor extends React.Component {
             return [
               <li
                 key={`dt-${idx}`}
-                className={classnames(name, "color", { selected: selectedColor === name })}
-                onClick={() => setSelectedColor({ name })}
+                className={classnames(name, "color", { selected: this.props.selectedColor && selectedColor === name })}
```
lmorchard: I don't think you need the `this.props.selectedColor &&` here - `selectedColor === name` should also turn up false if `selectedColor` is `null`

lmorchard:
Looks good! Thanks for working through the changes!

shr7:
@lmorchard Thanks for all the guidance. I really appreciate it :)

lmorchard:
46c6f4806bef4734984fc3e25b69b7d5961e461f

PR #188, opened 2018-03-21
make it respond

### LINKED ISSUE #41: Ensure responsive styles work down to 640px width
From [acceptance criteria](https://github.com/mozilla/ThemesRFun/blob/master/docs/acceptance.md):

> - [ ] It should be functionally responsive down to 640px.

Another final checklist item for near the end

johngruen:
Fixes #41


src/preset-themes/006.json
```diff
@@ -0,0 +1,44 @@
+{
```
pdehaan: Add: `"title": "006",`

src/preset-themes/default.json
```diff
@@ -32,6 +32,11 @@
       "h": 0,
       "s": 0,
       "l": 100
+    },
+    "tab_line": {
+      "h": 360,
+      "s": 100,
+      "l": 60
```
pdehaan: You'll need to tweak the /docs/theme-schema.json file as well. Along with all the other hsl[a] definitions, add the following new property:

```js
                "tab_line": {
                    "type": "object",
                    "properties": {
                        "h": {
                            "type": "integer",
                            "maximum": 360,
                            "minimum": 0
                        },
                        "s": {
                            "type": "integer",
                            "maximum": 100,
                            "minimum": 0
                        },
                        "l": {
                            "type": "integer",
                            "maximum": 100,
                            "minimum": 0
                        }
                    },
                    "required": ["h", "s", "l"],
                    "additionalProperties": false
                }
```

It's a bit of a funny one, since we've only added a `tab_line` definition to 2 of the 8 themes, so it isn't a required property unless we backfill it in to the other themes. But this should be enough to pass the super awesome theme linter.

johngruen:
Oh yeah,  i didn't see the new linter stuff after i rebased. looks awesome. I added tab_lines to all the current present themes (which are all going to change anyway)

pdehaan:
There is some cropping of the color pickers at 640 where the panel floats slightly off screen. Not sure if that is fixable, or even worth fixing. Stuff doesn't start _breaking_ until around 620px wide, where the "B" text input disappears off the right edge.

themer640

Similar stuff happens (to lesser degree) for the "Toolbar Icons and Text" color picker on the left side of the UI, were the "Hex" text input and color chips float slightly off the edge, but is still usable:

firefox_themer_640b

Probably the most problematic is the Theme Texture picker, since it causes some horizontal scrollage.

firefox_themer_theme_texture

firefox_themer_-dual_scrollage

johngruen:
@pdehaan yeah...maybe follow up bugs?

Issue #94, opened 2018-02-27
Enable users to contribute themes back to the project for inclusion in presets


lmorchard:
Just merged [the hotdog stand theme PR](https://github.com/mozilla/ThemesRFun/pull/90) from @6a68, and it occurs to me it might be interesting to have some mechanism to allow users to submit their favorite themes as contributions to our set of presets. 

Could do something that takes a shared theme URL and converts it to theme JSON to add to the project. Also had a notion to switch from the monolithic constants file to a directory of .json files that webpack assembles into an array for the editor

johngruen:
@pdehaan that's basically it...basically all these values get passed into CSS vars in Firefox and some get referenced in contexts that don't seem to allow alpha values

lmorchard:
Also, the only reason all the colors have alpha even where they're not supported is because, at one point, I didn't realize that - so all the presets just got alphas on all the things. They don't have to be there, I don't think

lmorchard:
Adding some more thoughts from another discussion:

We'd like to come up with a way to allow users to send us their favorite themes. Eventually, these themes will become new JSON files under src/preset-themes if we accept them for inclusion.

One way to do this would be to require that users open a Pull Request on GitHub with a JSON file provided by the Themer editor. However, this would limit theme contributors to people who know how to use git and GitHub.

We'd like to make it easier to submit themes back to us so that more people can participate. So, we'll have to get a little creative here. Maybe we could accept theme URLs in emails? Maybe we could set up a Google Docs form to accept theme URLs? Maybe we need to setup a server application somewhere to receive submissions directly from the editor application?

And then, on our side, we may need to build some tools to convert whatever we accept from users (e.g. theme URLs) into JSON files that we can put into src/preset-themes/. This could be a manual process, or we could make it more automatic to make it easier on us.

Issue #93, opened 2018-02-27
Generate a standalone theme add-on from template?


lmorchard:
Just an idea from IRC: What if we had an option to produce an actual standalone add-on from templated code that implemented a theme from the editor? Could be a real thing a user could submit to AMO, or possibly take it as a starting point for more advanced theme tinkering.

nt1m:
Yes, it would be really awesome to be able to publish to AMO easily.

lmorchard:
Copying some thoughts over from another discussion...

An interesting page of MDN doco to read on themes:

https://developer.mozilla.org/en-US/Add-ons/Themes/Theme_concepts

TL;DR: There are two kinds of themes we can make - static and dynamic. 

A static theme is pretty easy, since it's just basically taking the same properties we use in preset theme JSON files and creating a WebExtension manifest.json file based on it. 

A dynamic theme is a more complex WebExtension that includes JS code to make calls to the Theme API - this is actually how the add-on we include with Themer works.

For this issue, I think we'll aim for creating static theme manifest.json files. For bonus points, we could also generate the JS code for a dynamic theme as a starting point for someone who wants to tinker further.

nt1m:
Note: while static themes are supported in Firefox, AMO doesn't support them yet. Supported is being added this quarter, so it should be hopefully finished soon: https://github.com/mozilla/addons/issues/501

johngruen:
Looks like this is going to create server complexity but is maybe possible...I'll talk to mconca about this.

lmorchard:
In progress work toward this as an AWS Lambda function - https://github.com/Shruti9520/ServerlessAddon

PR #197, opened 2018-03-21
make debugger available on non-dev

johngruen:
@lmorchard this makes the debugger visible locally and hidden, but available everywhere else. Bumped the add-on version b/c i'd like to deploy to more easily extract all the themes i have on my github.io version.


lmorchard:
Looks fine to me - at some point, probably part of #94 and/or #93, we should rework this into an actual theme export feature

Issue #203, opened 2018-03-23
New name:  Firefox Color


wresuolc:
New approved name:  Firefox Color.  Please change anything as appropriate.  /cc @jbuck

jbuck:
@wresuolc Does this also mean the production domain should be changed to color.firefox.com ?

wresuolc:
@jbuck yes please. sorry for the trouble.  We discovered https://github.com/mjswensen/themer late in the game

lmorchard:
This should be done now

PR #216, opened 2018-04-02
Tweak the add-on ID and name based on env vars

### LINKED ISSUE #185: Allow multiple sites to be defined in SITE_URL
Right now we currently have SITE_URL=https://themer.stage.mozaws.net/ but we'll need to add https://themer.firefox.com to the list for production. The [web extension build process](https://github.com/mozilla/Themer/blob/e727ee21c9179b88fb954c21d3d3c30958c3ad65/webpack.extension.js#L54-L57) only allows a single URL.

I think the best way to do this might be doing something like `manifest.content_scripts[0].matches = process.env.SITE_URL.split()` but that's just me spit-balling.

lmorchard:
- New SITE_ID env var at build time will be added as a suffix to add-on
  name and ID

- If NODE_ENV is "development" the suffix will also appear with -dev

Fixes #185

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,10 +15,10 @@
   },
   "scripts": {
     "start": "npm-run-all --parallel server watch:extension watch:lint watch:validate",
-    "server": "cross-env NODE_ENV=development webpack-dev-server --config webpack.web.js",
+    "server": "cross-env NODE_ENV=development SITE_ID=local webpack-dev-server --config webpack.web.js",
     "watch": "npm-run-all --parallel watch:*",
-    "watch:web": "cross-env NODE_ENV=development webpack --watch --progress --colors --config webpack.web.js",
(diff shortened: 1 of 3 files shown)


lmorchard:
So, for local development, this should result in an add-on named "Themer (local-dev)" with ID "themer-local-dev@mozilla.com"

On github pages deployment, this should result in an add-on named "Themer (github)" with ID "themer-github@mozilla.com"

For staging, if we build with a SITE_ID=stage env var, it should end up as "Themer (staging)" and "themer-staging@mozilla.com"

For production, just not setting a SITE_ID and ensuring NODE_ENV=production should result in "Themer" and "themer@mozilla.com"

(Also, the "themer" name will change, but will handle that in issue #203)

capture

lmorchard:
Could hardcode this to watch for specific SITE_URL values, but felt like that might not be the best approach

PR #231, opened 2018-04-14
Dedupe saved themes

### LINKED ISSUE #166: Dedupe saved themes?
### Steps to reproduce:
1. Go to https://mozilla.github.io/Themer/
2. Click the [Save] button on your theme a ~~few~~ dozen times.

### Actual results:
A few dozen duplicate themes in your "Saved themes" section.

### Expected results:
Maybe we should disallow people to save a theme if they have the same hash already saved in their saved themes.

csd713:
It Fixes #166 
Disallows to save a theme, if it is present in the saved theme already.

@pdehaan @johngruen please review this.


src/web/lib/storage.js
```diff
@@ -17,16 +17,33 @@ const generateThemeKey = () =>
...
+  });
+}
+
+function checkDuplicateTheme(theme, callback) {
+  const themesList = listThemes();
+  let isThemeDuplicate = false;
+  for (const key of Object.keys(themesList)) {
+    isThemeDuplicate = themesEqual(theme, themesList[key].theme);
+    if (isThemeDuplicate) {
+      callback(isThemeDuplicate);
```
pdehaan: From the looks of this, wouldn't the callback be called twice here? Once on L42 if a match was found, and again on L46 after the `break` statement breaks out of the `for..of` loop.

src/web/lib/storage.js
```diff
@@ -17,16 +17,32 @@ const generateThemeKey = () =>
...
-  localStorage.setItem(
-    storageKey,
-    JSON.stringify({
-      theme,
-      modified: Date.now()
-    })
-  );
-  notifySelfForStorage(storageKey);
+  checkDuplicateTheme(theme, (isThemeDuplicate) => {
+    if (!isThemeDuplicate) {
```
meandavejustice: Since `localStorage.setItem` is syncronous, you don't need a callback here. You can just return the result from `isThemeDuplicate`

src/web/lib/storage.js
```diff
@@ -17,16 +17,32 @@ const generateThemeKey = () =>
...
+
+function checkDuplicateTheme(theme, callback) {
+  const themesList = listThemes();
+  let isThemeDuplicate = false;
+  for (const key of Object.keys(themesList)) {
+    isThemeDuplicate = themesEqual(theme, themesList[key].theme);
+    if (isThemeDuplicate) {
+      break;
+    }
+  }
```
meandavejustice: This could be written a little more cleanly, maybe something like:

``` js
return Object.keys(listThemes()).filter((key) => {
  return themesEqual(theme, themesList[key].theme);
}).length > 0;
```

pdehaan: @meandavejustice: Wouldn't we want something like [`Array#some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) [instead of `Array#filter()`] which should exit the loop as soon as a match is found?

meandavejustice: @pdehaan Sure thing, `some` is even better!

csd713: Thanks for this suggestion too. I learnt a new hack.

src/web/lib/storage.js
```diff
@@ -17,16 +17,26 @@ const generateThemeKey = () =>
...
+        modified: Date.now()
+      })
+    );
+    notifySelfForStorage(storageKey);
+  }
+}
+
+function checkDuplicateTheme(theme) {
+  const themesList = listThemes();
+  return Object.keys(listThemes()).some((key) => {
```
pdehaan: Sorry for being annoying...
Small nit, the `listThemes()` here should be `themesList` from L36 above, so we aren't calling the method twice.

csd713: Ahh, I missed that one. Thanks. Will push the changes.

src/web/lib/storage.js
```diff
@@ -17,16 +17,26 @@ const generateThemeKey = () =>
...
+    const storageKey = themeStorageKey(key);
+    localStorage.setItem(
+      storageKey,
+      JSON.stringify({
+        theme,
+        modified: Date.now()
+      })
+    );
+    notifySelfForStorage(storageKey);
+  }
```
pdehaan: Semi-related, but do we need an `else` statement?
Re-reading @johngruen's https://github.com/mozilla/FirefoxColor/issues/166#issuecomment-373050617, it sounded like we may want to pop the newly saved duplicate theme to the front of the list if it was a duplicate. Not sure if the code is currently doing that and I'm just not reading it correctly, or if we need an `else` where we update the `modified` date to `Date.now()` and do whatever animations we wanted.

csd713: Can we merge this code and open a new issue to include animation effect?

csd713: Hi @meandavejustice that sounds good.
While testing the changes after resolving the conflicts, I found that 'Save' button is disabled if there is no change from 'currentSaved' theme.
One case where this functionality (of checking duplicate saved themes) would be useful is when a user presses 'undo/redo' button and tries to save the same theme again.
So, would you still recommend to add this functionality?
Please let me know if I'm missing something here.

csd713: @pdehaan @meandavejustice Could you please review if we still need to fix this issue?
If not, we can close this PR.

lmorchard:
Cleaned up the conflicts and think I'm going to merge this. There's some UI polish we can add on top of this, but it works as a start!

PR #238, opened 2018-04-18
Switch to RGBA colors from HSLA colors

lmorchard:


src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -67,23 +68,24 @@ class ThemeColorsEditor extends React.Component {
...
+                />
+                <span className="color__label" title={colorLabels[name]}>
+                  {colorLabels[name]}
+                </span>
                 <span className="color__picker">
                   <SketchPicker
-                    color={{ h: color.h, s: color.s, l: color.l, a: color.a * 0.01 }}
+                    color={color}
                     disableAlpha={!colorsWithAlpha.includes(name)}
-                    onChangeComplete={({ hsl: { h, s, l, a } }) => {
```
lmorchard: All of this was super awkward and should have been refactored out awhile ago. Switching to RGBA seems to make this simpler.

src/web/lib/components/ThemeColorsEditor/index.js
```diff
@@ -36,10 +37,15 @@ class ThemeColorsEditor extends React.Component {
   handleKeyPress(event) {
     const { selectedColor, setSelectedColor } = this.props;
     if (event.keyCode === ESC && selectedColor !== null) {
-        setSelectedColor({ name: null });
+      setSelectedColor({ name: null });
     }
   }
 
+  handleColorChange(name, color) {
+    this.props.setColor({ name, color: color.rgb });
```
lmorchard: No color math anymore, just use the RGBA representation directly.

src/lib/themes.js
```diff
@@ -12,32 +15,39 @@ export const bgImages = require.context(
...
-    ? `hsl(${h}, ${s}%, ${l}%)`
-    : `hsla(${h}, ${s}%, ${l}%, ${a * 0.01})`;
+export const themesEqual = (themeA, themeB) =>
+  // HACK: "deep equal" via stringify
+  // http://www.mattzeunert.com/2016/01/28/javascript-deep-equal.html
+  JSON.stringify(themeA) === JSON.stringify(themeB);
+
+export const makeTinycolor = colorIn => {
+  let a = colorIn.a;
+  if (typeof a !== "undefined" && a > 1.0) {
```
lmorchard: This is mainly for backwards-compatibility with themes already in localstorage with the old HSLA format. We could drop this eventually.

src/lib/themes.js
```diff
@@ -12,32 +15,39 @@ export const bgImages = require.context(
...
+    // HACK: If the alpha channel is > 1.0 then assume it's a percentage that
+    // needs to be normalized to 0.0 - 1.0 range
+    a = Math.floor(a) / 100.0;
+  }
+  return tinycolor({...colorIn, a});
 };
 
+export const colorToCSS = colorIn => makeTinycolor(colorIn).toRgbString();
```
lmorchard: Turns out that tinycolor2 makes this simpler.

src/lib/metrics.js
```diff
@@ -53,8 +53,8 @@ const COLORS_TO_DIMENSIONS = {
   toolbar_field_text: "cd10"
 };
 
-const hslaToCSV = (name, { h, s, l, a }) =>
-  `${h},${s},${l}${colorsWithAlpha.includes(name) ? `,${a}` : ""}`;
+const rgbaToCSV = (name, { r, g, b, a }) =>
```
lmorchard: tinycolor2 didn't come with a util to generate this directly :/

package.json
```diff
@@ -115,7 +115,8 @@
     "redux-actions": "^2.2.1",
     "redux-devtools-extension": "^2.13.2",
     "redux-undo": "^1.0.0-beta9-9-7",
-    "testpilot-ga": "^0.3.0"
+    "testpilot-ga": "^0.3.0",
+    "tinycolor2": "^1.4.1"
```
lmorchard: FWIW, this is the library react-color uses under the hood to handle colors anyway, so it's handy to use it ourselves

bin/convert-themes-to-rgb.js
```diff
@@ -0,0 +1,25 @@
+const fs = require("fs");
```
lmorchard: This quick & dirty script is responsible for all the changes under `src/preset-themes` - I didn't manually edit any of them.

lmorchard:
Whoops - forgot to update the schema, update in a sec

bin/convert-themes-to-rgb.js
johngruen: oh good :)

johngruen:
Before
image

After
image

johngruen:
@lmorchard there seem to be some changes to the actual colors going on with the switch to RGB

src/preset-themes/008.json
```diff
@@ -2,44 +2,44 @@
   "title": "008",
   "colors": {
     "toolbar": {
-      "h": 120,
-      "s": 1,
-      "l": 86,
-      "a": 100
+      "r": 184,
+      "g": 255,
```
johngruen: So this color is way wring...i think something's happening where 1 is being converted to 100% saturation by tinycolor?

src/preset-themes/014.json
```diff
@@ -2,44 +2,46 @@
...
   "colors": {
     "toolbar": {
-      "h": 300,
-      "s": 87,
-      "l": 89,
-      "a": 86
+      "r": 251,
+      "g": 203,
+      "b": 251,
+      "a": 1
```
johngruen: The alpha values aren't being converted correctly

src/preset-themes/008.json
johngruen: I guess i commented the wrong line, the old saturation value here is 1 which may be getting converted to 100% somewhere in tinycolor before it gets converted to rgb?

src/preset-themes/001.json
```diff
@@ -2,44 +2,46 @@
...
+      "b": 227,
+      "a": 1
     },
     "toolbar_text": {
-      "h": 213,
-      "s": 43,
-      "l": 89
+      "r": 215,
+      "g": 226,
+      "b": 239
```
pdehaan: Would it make sense to convert these to HEX at this point and just do something like `"toolbar_text": "#2e3848"`,
and `"toolbar": "#433d2d"`.

I couldn't actually find any color values with a non "0" or non "1" value. Which now actually seems pretty suspicious.

src/preset-themes/001.json
```diff
@@ -2,44 +2,46 @@
...
     },
     "toolbar_field": {
-      "h": 211,
-      "s": 0,
-      "l": 100,
-      "a": 53.33333206176758
+      "r": 255,
+      "g": 255,
+      "b": 255,
+      "a": 1
```
pdehaan: I feel like this is weird and wrong. Shouldn't the "a" value here be something like "0.533"?

src/preset-themes/001.json
lmorchard: > Would it make sense to convert these to HEX at this point

Not really - trying to keep these JSON files very close to / exactly matching the internal representation of redux app state to keep things simple. And so far, that means separate r/g/b/a properties

src/preset-themes/003.json
```diff
@@ -2,44 +2,46 @@
...
   "colors": {
     "toolbar": {
-      "h": 200,
-      "s": 29,
-      "l": 32,
-      "a": 17
+      "r": 58,
+      "g": 89,
+      "b": 105,
+      "a": 1
```
pdehaan: Should this be 0.17? (trying to spot check some of these against http://hslpicker.com/)

src/preset-themes/006.json
```diff
@@ -2,44 +2,44 @@
...
     },
     "toolbar_field": {
-      "h": 260,
-      "s": 74,
-      "l": 51,
-      "a": 11
+      "r": 99,
+      "g": 38,
+      "b": 223,
+      "a": 1
```
pdehaan: `"a": 0.11`?

bin/convert-themes-to-rgb.js
```diff
@@ -0,0 +1,25 @@
...
+    Object.entries(theme.colors).forEach(([name, color]) => {
+      const rgba = tinycolor(color).toRgb();
+      if (!colorsWithAlpha.includes(name)) {
+        delete rgba.a;
+      }
+      theme.colors[name] = rgba;
+    });
+    fs.writeFileSync(
+      path.join(themesPath, filename),
+      JSON.stringify(theme, null, "  "),
```
pdehaan: Nit: we'll need to add a `\n` at the end here if we don't want to lose our trailing blank line in those files. 🤷‍♀️

src/preset-themes/008.json
pdehaan: Yeah, according to http://hslpicker.com/#dbdcdb, this should be `{r: 219, g: 220, b: 219, a: 1}`

bin/convert-themes-to-rgb.js
lmorchard: Not sure this is important

bin/convert-themes-to-rgb.js
```diff
@@ -0,0 +1,35 @@
+const fs = require("fs");
+const path = require("path");
+const tinycolor = require("tinycolor2");
+const colorsWithAlpha = ["toolbar", "toolbar_field"];
+const themesPath = path.join(__dirname, "..", "src", "preset-themes");
+
+const makeTinycolor = colorIn => {
```
lmorchard: Copied over the tweak to alpha conversion I was missing from the themes library. That should fix issues with alpha values

lmorchard:
Okay, a couple of tweaks and I think I've got the alpha & saturation issues fixed. I don't have a great eye for the color differences, but I *think* that gets things matching the "Before" image up there again?

lmorchard:
Another quick tweak to fix linting errors. Also FWIW, the alpha & saturation issues weren't caused by tinycolor - I just missed carrying over the hacks multiplying values by 100 to try to avoid JS floating point issues. Post-conversion, tinycolor should be reliable since we're not going to be dealing with that math anymore

johngruen:
Looks great!

image

PR #243, opened 2018-04-19
184 update save btn ux

### LINKED ISSUE #184: Change Save button text to Saved when it is clicked.
@lmorchard @pdehaan 
When I click on copy button it is changed from `copy` to `copied`. Which gave an excellent user experience. But when I click on `Save` button nothing happens. After clicking on save button, themes got saved down the page. So, for a new user, it takes some time to realize what is actually happening. I think it would be better if we change the save button text from `save` to `saved`, once it is clicked. 

Before clicking `Copy` button-
themerissue1
After clicking `Copy` button-
themerissue2

But in case of `Save` button-
Before clicking-
themerissue3

After clicking-
themerissue4

Thanks!

HandsomeMedia:
Fixes #184
- initial state shows disabled button with "Save" text
- edited state shows enabled button with "Save" text
- saved state shows disabled button with "Saved!" text

### DIFF
--- a/src/lib/store.js
+++ b/src/lib/store.js
@@ -25,7 +25,8 @@ export const actions = {
     "CLEAR_PENDING_THEME",
     "SET_LOADER_DELAY_EXPIRED",
     "SET_SAVED_THEMES",
-    "SET_SAVED_THEMES_PAGE"
+    "SET_SAVED_THEMES_PAGE",
+    "SET_CURRENT_SAVED_THEME"
   ),
   theme: {
     ...createActions({}, "SET_THEME", "SET_COLOR", "SET_BACKGROUND"),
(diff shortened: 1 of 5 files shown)


lmorchard:
Works for me!

HandsomeMedia:
Hah! Well I'm astounded there are no issues with best practice, given my novice React skills...  @lmorchard 
 - If you have a moment, could you or another maintainer give it a once over - surely I done goofed somewhere?

lmorchard:
> Hah! Well I'm astounded there are no issues with best practice, given my novice React skills

It all looks pretty straightforward and clean to me! I was trying to think of a simpler way to determine if the last-saved theme had changed, rather than keeping a copy around in UI state. But, I couldn't think of anything better so 🤷‍♂️ :shipit:

HandsomeMedia:
I will pat myself on the back then.  Thanks!

Issue #169, opened 2018-03-14
Install Themer banner is briefly displayed when the page is refreshed


SoftVision-CosminMuntean:
**[Affected versions]:**
- Firefox 58.0.2 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Themer" add-on (0.0.30 custom build) installed from the "about:debugging" page.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Click the "Themer" toolbar button.
3. Refresh the page and observe the behavior.

**[Expected result]:**
- The page is correctly refreshed.

**[Actual result]:**
- When the page is refreshed, the install themer banner is briefly displayed at the top of the page.

**[Notes]:**
- Attached a screen recording of the issue: 
![install banner](https://user-images.githubusercontent.com/19226534/37405780-3ba5a380-279e-11e8-867e-f3d5cc754e27.gif)

johngruen:
Yeah i've been thinking about this one, and we might be best served by not having this banner move page content up or down so that even if it flashes the content doesn't jump

lmorchard:
This is also where the loader is kind of a guess - could be fixed by increasing the loader wait, since this stuff is unpredictable based on performance where it's run

Softvision-AndreeaCupsa:
This issue is still reproducible on latest Firefox (59.0.3) and latest Nightly (61.0a1) with Firefox Color(0.0.35) installed, on Win10 x64 and Mac 10.13.3. When refreshing the page "Install Themer" banner is briefly displayed.

johngruen:
I have a feeling this is because we haven't done a deployment recently.

lmorchard:
Well, a deployment won't affect localhost either way. Probably still just going to show this banner on some machines unless we keep bumping the loader delay up

Softvision-AndreeaCupsa:
This issue is still reproducible on Win 10 x64, Mac 10.13.3 and Arch Linux, using latest Firefox (59.0.3) and latest Nightly (61.0a1) with Firefox Color (0.0.36) installed. The banner is still briefly displayed when refreshing the page and also when you navigate to Firefox Color site.

lmorchard:
Unassigning myself for now, because I'm not actively working on it at the moment. Will probably get back to looking, but didn't want to block

Softvision-AndreeaCupsa:
[...]

PR #245, opened 2018-04-19
Loader tweaks

### LINKED ISSUE #209: Some hiccups with the loading screen
Two issues: 
1. The loader should have a minimum display time of 2 seconds
2. The loader seems to be rendering twice which causes a flash of text. It should only render once.

lmorchard:
- Increase base loader delay period to 2 seconds

- Select quote once at page init, rather than for every app state render

Fixes #209


src/web/lib/components/AppLoadingIndicator/index.js
```diff
@@ -1,15 +1,13 @@
 import React from "react";
 import "./index.scss";
-import { loaderQuotes } from "../../../../lib/constants";
 
-export const AppLoadingIndicator = () => {
-  const i = Math.floor(Math.random() * loaderQuotes.length);
```
lmorchard: Turns out this was selecting a new quote every time the app renders - which is every time the state changes. And the state changes several times during loading, which is why we have a spinner :)

src/web/index.js
```diff
@@ -30,7 +30,7 @@ const clipboard = new Clipboard(".clipboardButton");
 const addonUrl = process.env.ADDON_URL;
 
 // Period after which app loading indicator will disappear if add-on not found
-const LOADER_DELAY_PERIOD = 500;
+const LOADER_DELAY_PERIOD = 2000;
```
lmorchard: This should give us a minimum of 2 seconds for the loader, but retaining the timer extensions every time state changes during initialization

lmorchard:
Also expecting this to fix #169, since we increase the loader delay and should better hide the state transition

PR #252, opened 2018-04-24
Minimal mobile

johngruen:
I did this on top of the terms and privacy PR, so we should probably merge that first and i an rebase this


johngruen:
Fixes #45

src/web/lib/components/App/index.js
```diff
@@ -1,20 +1,21 @@
-import React from "react";
+import React, { Fragment } from "react";
```
johngruen: Fragments are pretty new, but super handy...https://reactjs.org/docs/fragments.html#short-syntax
Basically let you return unwrapped components.

johngruen:
This branch adds mobile UI and cleans up the  `App` component some more

PR #260, opened 2018-05-01
Set cd2 metric flag when the site is visited from add-on

### LINKED ISSUE #257: cd2 is not working as described
`cd2` captures if a user begins a visit to the site from the browserAction, but it doesn't seem to be firing...

STR
1. Open the site from the browserAction
2. Pick a new preset theme
3. In the console, look at the metrics log object...`cd2` is set to `false`

expected: `cd2` is set to `true`

lmorchard:
Fixes #257


johngruen:
👍

Issue #237, opened 2018-04-18
The "Feedback" button is not expanded when focused using the keyboard


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 59.0.2 and above

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest version of the "Firefox Color" add-on (0.0.35 custom built on 18.04.2018) installed.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites and click the Firefox Color toolbar button.
2. Click somewhere in the "Firefox Color" string.
3. Press the "Tab" key twice and observe the behavior.

[Expected result]:
- The button expands.

[Actual result]:
- The button is focused but it does not expand.

[Notes]:
- Attached a screen recording of the issue:
![feedback button is not expanded when focused via keyboard](https://user-images.githubusercontent.com/18304511/38939345-680ba688-4330-11e8-8043-9af139cc74b3.gif)

dominictwlee:
Hi, I'm a newbie contributor to open source but I could give this a go if no one else is working on it.

Softvision-AndreeaCupsa:
I have verified that the issue is no longer reproducible on Mac 10.13.3, Win10 x64 and Arch Linux on latest Firefox (59.0.3) and latest Nightly (62.0a1) with Firefox Color (0.0.36) installed.
The "Feedback" button is expanded when focused using the keyboard.

PR #262, opened 2018-05-02
Expand feedback button when focused using a keyboard (#237)

dominictwlee:
I've made the feedback button expandable when accessed using a keyboard.


johngruen:
Nice...could you add focus states to the color selector as well...that's here: https://github.com/dominictwlee/FirefoxColor/blob/master/src/web/lib/_common.scss#L53

There are actually three button types which could use :focus added

dominictwlee:
Sure, I'll look into it, thanks for the feedback.

PR #272, opened 2018-05-07
touch up header UI

johngruen:
Minor style tweaks to header UI


src/web/lib/components/AppHeader/index.js
```diff
@@ -23,6 +25,7 @@ export const AppHeader = ({ hasExtension }) => {
               onClick={() => Metrics.linkClick("test-pilot")}
               target="_blank"
               rel="noopener noreferrer"
+              style={{ color: highlightColor }}
```
lmorchard: Works for me - but FWIW, the contrast on "Firefox Test Pilot" and the background looks really unreadable to me personally across many of the preset themes.

PR #274, opened 2018-05-08
Fix #270.  Add an update URL

### LINKED ISSUE #270: Add an update URL?
Saw https://github.com/mozilla/side-view/pull/232 and wanted to confirm we are 👌 for this as well.

I ran <kbd>$ npm run build:extension</kbd> but didn't see an `applications.gecko.update_url` in build/extension/manifest.json:

```js
  "applications": {
    "gecko": {
      "id": "FirefoxColor@mozilla.com",
      "strict_min_version": "57.0a1"
    }
  },
```

---

```sh
$ git rev-parse --short HEAD # 0de7d69
$ npm run build:extension
$ jq -r ".applications.gecko" build/extension/manifest.json
{
  "id": "FirefoxColor@mozilla.com",
  "strict_min_version": "57.0a1"
}
```

clouserw:
This should fix #270 but the project still isn't hosted on Test Pilot yet so the URLs haven't been verified.  We can wait to land if you want, or land and we'll check later. 

pinging @jbuck for status update on that side

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "lint:css": "stylelint 'src/**/*.scss'",
     "lint:js": "eslint --color .",
     "validate": "npm-run-all validate:*",
-    "validate:extension": "addons-linter build/extension",
+    "validate:extension": "addons-linter build/extension --self-hosted",
     "validate:themes": "node ./bin/validate-themes.js",
     "test": "npm run --parallel lint validate",
     "deploy": "gh-pages -x -d build/web -r \"https://$GH_TOKEN@github.com/mozilla/FirefoxColor.git\"",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -157,7 +157,8 @@
     "applications": {
       "gecko": {
         "id": "FirefoxColor@mozilla.com",
-        "strict_min_version": "57.0a1"
+        "strict_min_version": "57.0a1",
+        "update_url": "https://testpilot.firefox.com/files/FirefoxColor@mozilla.com/updates.json"
```
lmorchard: I think we might need to play with this URL on a per-build basis: The add-on is built specifically for each environment - i.e. prod, stage, dev, local - and so the update URL won't be the same for each

lmorchard:
Actually, I think I'm going to merge this as-is because it *does* work for prod by default, which is correct. We can add tweaks in the build script in a separate PR

PR #284, opened 2018-05-11
update banners

johngruen:


src/web/lib/_picker.scss
```diff
@@ -23,7 +23,7 @@
     border-left: 12px solid transparent;
     border-right: 12px solid transparent;
     border-top: 12px solid var(--white);
-    bottom: -12px;
+    bottom: -11px;
```
johngruen: small adjustments here make these more zoom friendly

PR #290, opened 2018-05-16
add twitter and og images

### LINKED ISSUE #174: The 'og:image' property should be explicitly provided
### PREVIEW:

og-preview

---

Via https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fmozilla.github.io%2FThemer%2F

og-themer

We'll also want to change that `og:url` once we're set up on some themer.firefox.com domain, or whatever.

johngruen:
Fixes #174


webpack.web.js
```diff
@@ -32,6 +32,8 @@ module.exports = merge(common.webpackConfig, {
       template: "./src/web/index.html.ejs",
       filename: "index.html",
       chunks: ["index"],
+      ogImage: `${packageMeta.homepage}images/color-fb.jpg`,
+      twitterImage: `${packageMeta.homepage}images/color-twitter.jpg`,
```
pdehaan: Does `packageMeta.homepage` get overridden per env, or will we need to update it to the prod URL before general release?

https://github.com/mozilla/FirefoxColor/blob/2852096f6e30627035c2ab852b74b719911d893b/package.json#L11

lmorchard: This should probably use `process.env.SITE_URL` instead of packageMeta. That *does* get changed per environment

johngruen: okay cool

PR #295, opened 2018-05-21
Onboarding WIP

johngruen:


src/extension/background.js
```diff
@@ -62,7 +62,7 @@ const storeTheme = ({ theme }) => browser.storage.local.set({ theme });
 const applyTheme = ({ theme }) => {
   log("applyTheme", theme);
   if (!theme) {
-    return;
```
johngruen: So if there's no theme when applyTheme is called we just open a Color Tab ... this should handle the case where the user installs from Test Pilot

src/lib/metrics.js
```diff
@@ -177,12 +177,13 @@ const Metrics = {
   installSuccess() {
     if (!installTimer) {
       // Skip this event if there's no install timer in progress.
-      return;
```
johngruen: using the installSuccess metrics function to return a boolean that will toggle an Oboarding component

src/web/index.js
```diff
@@ -107,7 +109,9 @@ window.addEventListener("message", ({ source, data: message }) => {
       const hasExtension = selectors.hasExtension(store.getState());
       if (!hasExtension) {
         store.dispatch(actions.ui.setHasExtension({ hasExtension: true }));
-        Metrics.installSuccess();
+        if (Metrics.installSuccess()) {
+          store.dispatch(actions.ui.setJustGotExtension({ justGotExtension: true }));
```
johngruen: Dispatch onboarding if that installSuccess event returns true

src/web/index.js
```diff
@@ -40,6 +41,7 @@ const jsonCodec = JsonUrl("lzma");
 
 const urlEncodeTheme = theme =>
   jsonCodec.compress(theme).then(value => {
+    console.log(theme); // eslint-disable-line
```
lmorchard: Use the `log()` util function for stuff like this - it doesn't need eslint-disable-line and the output gets turned off in production builds

src/lib/metrics.js
lmorchard: Probably shouldn't try hooking into this - it's just a timer for reporting to metrics if the add-on got installed within a reasonable amount of time.

lmorchard:
Okay, pushed a commit to add a `?firstRun=1` param to the web app. The add-on will launch a tab to that on the first startup, then set a flag to skip on subsequent startups. You should be able to preview the onboarding by just manually adding the `?firstRun=1` param

johngruen:
@lmorchard problem i'm seeing is that there's no first run if the add-on is installed from the Firefox Color site

lmorchard:
You mean from your own local copy of the site?

lmorchard:
Okay, this got more complicated than I expected, but I *think* I've got it worked out except for a flash of default theme during add-on install that should flip back to the last-selected theme when things all settle down

lmorchard:
Going to merge and we can keep tweaking from there

Issue #208, opened 2018-03-27
Add an onboarding experience


johngruen:
The site does not provide much in the way of scaffolding users right now. It would be good to add some sort of minimal on-boarding experience. We could probably do this by embedding a YouTube video or something in the `Banner` component.

I'll work with Sevaan and Cherry to figure out what the right UX would be here.

PR #316, opened 2018-05-31
add final onboarding content

johngruen:
Resolves onboarding


johngruen:
Fixes #208

lmorchard:
Would be nice to have a way to bail out of onboarding early and maybe go back after hitting next. But, merging regardless.

PR #317, opened 2018-05-31
update a few presets

johngruen:


src/preset-themes/002.json
```diff
@@ -8,35 +8,39 @@
...
+      "a": 1
     },
     "accentcolor": {
-      "r": 9,
-      "g": 43,
-      "b": 78
+      "r": 14,
+      "g": 41,
+      "b": 65,
+      "a": 1
```
pdehaan: Theme validation failed for 002.json:
- `data.colors.accentcolor` should NOT have additional properties,
- `data.colors.toolbar_field_text` should NOT have additional properties,
- `data.colors.toolbar_text` should NOT have additional properties,
- `data.colors.tab_line` should NOT have additional properties

I think it's failing because of the extra `a` (alpha) values.

PR #319, opened 2018-05-31
Rework themesEqual() to use fuzzy equality on alpha channels  

lmorchard:
JS FP WTF FML

### DIFF
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -1,13 +1,79 @@
 import tinycolor from "tinycolor2";
-import { colorsWithAlpha } from "./constants";
+import { colorsWithAlpha, alphaEqualityTolerance } from "./constants";
 import { presetThemesContext, bgImages } from "./assets";
 
 const defaultTheme = presetThemesContext("./default.json");
 
-export const themesEqual = (themeA, themeB) =>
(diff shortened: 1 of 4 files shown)


lmorchard:
Or in other words, I broke down and rewrote `themesEqual()` to be more explicit than a magical JSON string compare and added a fuzz factor to alpha channels because apparently `0.86` is sometimes `0.8599999999901` or some nonsense

johngruen:
Okay looking at this now

__takes belt of whiskey__

src/lib/themes.js
```diff
@@ -1,13 +1,79 @@
...
-  // HACK: "deep equal" via stringify
-  // http://www.mattzeunert.com/2016/01/28/javascript-deep-equal.html
-  JSON.stringify(themeA) === JSON.stringify(themeB);
+export const themesEqual = (themeA, themeB) => {
+  if (!!themeA !== !!themeB) {
+    return false;
+  }
+
+  const hasImagesA =
+    "images" in themeA && "additional_backgrounds" in themeA.images;
```
johngruen: TIL you can do this

johngruen:
works/makes sense!

johngruen:
@lmorchard shi

Issue #311, opened 2018-05-30
The "Save" button is wrongly active on all saved themes except the last


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 60.0.1 and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a new Firefox profile with the latest version of the "Firefox Color" add-on (0.0.39 custom built on 30.05.2018) installed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites and save at least two themes.
2. Click on each saved theme from the "Saved themes" area and observe the "Save" button status.

**[Expected result]:**
- The "Save" button is deactivated on each saved theme.

**[Actual result]:**
- The "Save" button is deactivated only on the last saved theme and it is active on the other ones.

**[Notes]:**
- Attached a screen recording of the issue:
![rec of the issue](https://user-images.githubusercontent.com/31243259/40718395-7804034c-6418-11e8-918b-2237aa02008e.gif)

Softvision-MariusComan:
This issue is still reproducible with the latest version of the "Firefox Color" add-on (2.0.0) installed from Test Pilot.

rebmullin:
The ui has changed but there was a similar issue with the new UI (see #886). This  should be fixed on stage. Please reopen if you see something different.

SoftVision-SimonaRosu:
I have verified this issue on the stage server of the Firefox Color website and is no longer reproducible. Tested using the latest Nightly 77.0a1 (Build ID: 20200504093644 ), Firefox RC 76 (Build ID: 20200427162639), Firefox Release 75 (Build ID:) on Windows 10 x64, Ubuntu 18.04 x64, Mac 10.14.6.
- The ”Save” button is changed to “Saved!” grayed out button when a theme is saved. 

However, the issue is still reproducible with the production version. Can you please confirm if the fix wasn’t deployed to prod?

caitmuenster:
Thanks @SoftVision-SimonaRosu! Correct, this hasn't been deployed to prod yet. Glad to know it's fixed on stage.

PR #320, opened 2018-05-31
update save behavior

johngruen:


johngruen:
Fixes #311

lmorchard:
Hmm, this disables adding an entry to the undo stack when picking a saved theme. But it doesn't seem to fix #311 for me - when I pick a saved theme, the "Save" button is still active

src/lib/store.js
```diff
@@ -115,6 +115,8 @@ export const reducers = {
       }),
       SET_THEME: (state, { meta }) => ({
         ...state,
+        // flush pending themes on save so that saved themes don't trigger modal
+        pendingTheme: null,
```
lmorchard: This should probably get handled by dispatching a `clearPendingTheme` action wherever necessary, rather than a side effect here

src/web/lib/components/SavedThemeSelector/index.js
```diff
@@ -21,7 +21,7 @@ export const SavedThemeSelector = ({
       themes={sortedSavedThemes}
       className="saved-theme-selector"
       previewClassName="saved-theme-preview"
-      onClick={theme => setTheme({ theme })}
+      onClick={theme => setTheme({ theme, userEdit: false })}
```
lmorchard: Making `meta.userEdit` false here gets this edit filtered out of the undo stack, but I still see the Save button get active when selecting a saved theme

lmorchard:
Feels like overkill, but to really fix this I think we need to stick a loop in ThemeBuilder that compares the current theme to every saved theme on render to decide whether or not the Save button is disabled.

johngruen:
@lmorchard thanks for the review. i can probably update tomorrow.

lmorchard:
Going to close this for now to help clean up the PR queue, we can take another swing at #311

PR #332, opened 2018-06-04
Send finish-visit metric event from add-on when tab closed or navigates away

### LINKED ISSUE #322: Cannot make finish-visit fire
finish-visit does not appear to be firing as intended. I see no record of it on the GA event logs.

lmorchard:
- Try to keep metrics in sync between tabs and add-on

- When site tab closes or navigates away, send the finish-visit metric event

Fixes #322

### DIFF
--- a/src/extension/background.js
+++ b/src/extension/background.js
@@ -2,9 +2,10 @@ import { makeLog } from "../lib/utils";
 import {
   normalizeTheme,
   normalizeThemeBackground,
-  colorToCSS,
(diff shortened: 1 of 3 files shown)


lmorchard:
Ended up being a bit more complicated than I expected, but I think this should get finish-visit firing

johngruen:
@lmorchard everything seems to be in order here...can you do a version bump then merge this to prod?

PR #339, opened 2018-06-05
Set cd1 metric when add-on installed / uninstalled

lmorchard:
So I saw [this issue over on Test Pilot](https://github.com/mozilla/testpilot/issues/3612) and mistook it for a Firefox Color issue. Turns out we aren't setting the add-on flag for metrics here, either. Doh.


lmorchard:
This has lingered awhile, going to self-merge.

Issue #336, opened 2018-06-05
Favicon attempting to load from localhost


jbuck:
```
$ curl https://color.firefox.com/
<!DOCTYPE html>
<html lang="en" prefix="og: http://ogp.me/ns#">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width" />
    <meta name="description" content="Theming experiment for Firefox Quantum and beyond." />
    <!-- Open Graph meta tags-->
    <meta property="og:title" content="Firefox Color" />
    <meta property="og:url" content="https://mozilla.github.io/FirefoxColor/" />
    <meta property="og:type" content="website" />
    <meta property="og:image" content="http://localhost:8080/images/color-fb.jpg" />
    <!-- Twitter card meta tags-->
    <meta name="twitter:card" content="summary" />
    <meta name="twitter:site" content="@FxTestPilot" />
    <meta name="twitter:title" content="Firefox Color" />
    <meta name="twitter:description" content="Theming experiment for Firefox Quantum and beyond." />
    <meta name="twitter:image" content="http://localhost:8080/images/color-twitter.jpg" />
    <title>Firefox Color</title>
    <link rel="icon" type="image/svg+xml" href="images/icon.svg" />
  <link href="index.css" rel="stylesheet"></head>
  <body>
    <div id="root"></div>
    <script type="text/javascript" src="./vendor/json-url.js"></script>
  <script type="text/javascript" src="index.js"></script></body>
</html>
```

lmorchard:
Hmm, this doesn't happen on the [development site](https://mozilla.github.io/FirefoxColor/). 

What `SITE_URL` and `SITE_ID` env var do we use for building the site in production? If missing, [`SITE_URL` will default to localhost:8080](https://github.com/mozilla/FirefoxColor/blob/master/webpack.common.js#L18)

lmorchard:
[Looks like it's happening on stage, too.](https://color.stage.mozaws.net/)

For production, I think we should have `SITE_ID=production SITE_URL=https://color.firefox.com/`

For stage, we should have `SITE_ID=stage SITE_URL=https://color.stage.mozaws.net/`

jbuck:
Okay, I've got [a PR](https://github.com/mozilla-services/cloudops-deployment/pull/2087/files) up to fix this in Jenkins. Trying stage build now

lmorchard:
[...]

PR #340, opened 2018-06-05
Change homepage to site URL to correct og:url

lmorchard:
Issue #336


lmorchard:
This has lingered awhile, going to self-merge.

Issue #309, opened 2018-05-29
The "https://mozilla.github.io/FirefoxColor/" URL is wrongly displayed instead of "https://color.firefox.com" in "about:addons"


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 60.0.1 and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a new Firefox profile with the latest version of the "Firefox Color" add-on (0.0.39 custom built on 29.05.2018) installed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Go to "about:addons#extensions" and click the "Fx Color" add-on's "More" button.
3. Observe the URL displayed in the "Homepage" field.

**[Expected result]:**
- The "https://color.firefox.com" URL is displayed.

**[Actual result]:**
- The "https://mozilla.github.io/FirefoxColor/" URL is displayed.

**[Notes]:**
- Attached a screenshot of the issue:
![color homepage ss](https://user-images.githubusercontent.com/31243259/40668417-214ab204-636d-11e8-8d42-803dfb4572b3.png)

ranjithrajv:
I changed the URL, can we close this issue.

lmorchard:
I can confirm the URL is changed. The URL is hardcoded right now, so an enhancement might be to change the URL to reflect which site the add-on is built to use (e.g. dev vs stage vs prod). But, in literal terms, I think this issue is fixed.

![capture](https://user-images.githubusercontent.com/21687/45230626-6b038a00-b297-11e8-861b-28244a6185a2.PNG)

PR #370, opened 2018-07-05
replaced URLs for homepage

ranjithrajv:
The "https://mozilla.github.io/FirefoxColor/" URL is wrongly displayed instead of "https://color.firefox.com" in "about:addons",  so replaced it.

https://github.com/mozilla/FirefoxColor/issues/309


lmorchard:
Thanks for the patch!

PR #372, opened 2018-07-07
Randomize presets

### LINKED ISSUE #366: Randomize order of preset themes
I feel like randomizing the order of the preset themes will add a little freshness everytime the user goes into Firefox Color.

I'm starting to get a little blindness to them, and I think randomizing the order will help keep things dynamic.

a8t:
Fixes #366: Randomize presets on first load.

Added a function to `lib/utils.js` that copies the array, shuffles it, and returns the shuffled copy.

Please LMK any thoughts!

### DIFF
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -1,6 +1,7 @@
 import tinycolor from "tinycolor2";
 import { colorsWithAlpha, alphaEqualityTolerance } from "./constants";
 import { presetThemesContext, bgImages } from "./assets";
+import { shuffle } from "./utils";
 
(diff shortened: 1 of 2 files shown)


lmorchard:
Thanks for the contribution! Sorry for the delay in merging - I think we're still considering whether we really do want to randomize presets. I'm tempted to merge this and then add a config switch

johngruen:
Hi a8t, thanks for your contribution. I think i am going to close this now, since we're currently using the array index of the preset themes in the test pilot metrics to keep an eye on the relative popularity of themes. shuffling the order will break these metrics. When we graduate from test pilot, we will consider revisiting this work!

PR #375, opened 2018-07-22
#Fixes 249 Mouse hover over heading

ritikavnair:
The mouse pointer now changes into a text select pointer on hover over "Share your theme" and the text below it, i.e., "Copy and paste this URL to share your creation."


lmorchard:
Thanks for the contribution!

PR #378, opened 2018-07-23
Add start:ec2 task that auto-sets host name from AWS API

lmorchard:
This ends up being handy for me to run a dev box in EC2, as well as setting HOST env var along with PORT to override defaults.


lmorchard:
Going to self-merge this if/when it passes build.

PR #391, opened 2018-08-07
Custom background images support

lmorchard:
Mostly functional, but some known things left to do, roughly in a priority order that we can probably put off doing about halfway through:

- [x] Rework process to add a new image such that it pops up a file selector immediately, rather than adding an empty row with a browse button - refactor image import button out into its own component
- [x] Ensure duplicate theme comparison logic takes custom images into account (i.e. two themes with identical colors but different custom images should compare as different)
- [x] [Implement new UX design!](https://mozilla.github.io/testpilot-assets/ThemesRFun/01_Measurements/Updated_FF_Color/#artboard4) (roughly done)
- [x] Add a ~~good~~ display of import validation errors
- [x] Figure out why sometimes reloading the page wipes the current theme (maybe fixed? can no longer reproduce)
- [x] Fix text highlighting bug when dragging image order
- [x] Finish reworking localStorage handling - also ensure the whole set of imported images isn't re-loaded from localStorage on every new image
- [x] Stop generating `?theme` URL param altogether when custom images present, not just theme with colors but backgrounds omitted
- ~~Tweak browser preview to show custom images?~~
- ~~Maybe move tiling & position options into constants.js?~~
- [x] Further tweak max image size and content types? (we can continue to iterate here)
- [x] Replace single awkward `setCustomBackground` Redux action with `addCustomBackground` & `updateCustomBackground` actions?

And as a final step, I plan to squash and clean up all these PROGRESS commits before moving to merge this beast.


lmorchard:
Hmm, I updated my branch, but this PR seems not to be reflecting latest changes. Might close this and reopen a fresh one

package.json
```diff
@@ -20,6 +20,7 @@
   },
   "scripts": {
     "start": "cross-env NODE_ENV=development SITE_ID=local SITE_HOST=localhost npm run start:common",
+    "start:mini": "cross-env NODE_ENV=development SITE_ID=local SITE_HOST=localhost npm-run-all --parallel server watch:extension",
```
lmorchard: Don't absolutely need this, but it came in handy for saving some CPU on my machine when I just wanted to build the site and not also lint & test on every change.

package.json
```diff
@@ -103,6 +104,7 @@
     "npm-run-all": "^4.1.2",
     "onchange": "^4.0.0",
     "prettier": "^1.13.7",
+    "react": "^16.4.2",
```
lmorchard: Mainly upgraded React here for [the new ref API](https://reactjs.org/docs/refs-and-the-dom.html), but didn't really end up needing it. Still not a bad idea though, seems not to break things.

src/extension/background.js
```diff
@@ -78,30 +85,54 @@ const finishVisit = tabId => {
...
-      siteTabMetrics[tabId] = message.params;
-      log("setMetrics for tab", tabId);
-      break;
-    case "ping":
-      port.postMessage({ type: "pong" });
-      break;
-    default:
-      log("unexpected message", message);
+  const { type } = message;
+  messageHandlers[type in messageHandlers ? type : "default"](message, port);
```
lmorchard: Moved away from a switch statement to a function dispatch table, because the local variable situation started getting hairy.

src/lib/store-test.js
```diff
@@ -0,0 +1,222 @@
+const { expect } = require("chai");
```
lmorchard: This is not a comprehensive suite of tests - I mostly just started doing TDD in the process of keeping things straight for this new feature.

src/lib/store.js
```diff
@@ -33,11 +37,30 @@ export const actions = {
...
+      "UPDATE_CUSTOM_BACKGROUND",
+      "CLEAR_CUSTOM_BACKGROUND",
+      "MOVE_CUSTOM_BACKGROUND"
+    ),
     // HACK: Seems like redux-undo doesn't have sub-tree specific undo/redo
     // actions - but let's fake it for now.
     undo: ActionCreators.undo,
     redo: ActionCreators.redo
+  },
+  images: {
```
lmorchard: Images are managed separately from themes so that themes can just refer to them by name, share the data between themes, and not have to shuttle the image blobs around during common redux reducer activity

src/lib/store.js
```diff
@@ -139,6 +174,42 @@ export const reducers = {
...
+          [name]: { name, importing, size, image, type }
+        }
+      }),
+      DELETE_IMAGES: (state, { payload: names = [] }) => {
+        const images = { ...state.images };
+        names.forEach(name => delete images[name]);
+        return { ...state, images };
+      },
+      LOAD_IMAGES: (state, { payload: images = {} }) => ({ ...state, images }),
+      // LOAD_IMAGE differs from UPDATE_IMAGE in that the middleware does not
```
lmorchard: Kind of a hack to reuse some common code - LOAD_IMAGE is for a local load of an image on localStorage changes between tabs, UPDATE_IMAGE is a more general action that also triggers updates in the add-on

src/web/index.html.ejs
```diff
@@ -19,6 +19,7 @@
     <link rel="icon" type="image/svg+xml" href="images/icon.svg" />
   </head>
   <body>
+    <div id="modal"></div>
```
lmorchard: Not super happy adding another root-level element, but this seemed the easiest (intended?) usage of React portals

src/web/index.js
```diff
@@ -38,11 +38,15 @@ let outstandingPings = 0;
...
 
-const urlEncodeTheme = theme =>
-  jsonCodec.compress(theme).then(value => {
-    const { protocol, host, pathname } = window.location;
-    return `${protocol}//${host}${pathname}?theme=${value}`;
-  });
+const urlEncodeTheme = ({ hasCustomBackgrounds = false, theme }) => {
+  const { protocol, host, pathname } = window.location;
+  const baseUrl = `${protocol}//${host}${pathname}`;
+  return hasCustomBackgrounds
```
lmorchard: Themes with custom background skip adding a parameter to the URL altogether

src/web/index.js
```diff
@@ -52,35 +56,13 @@ const postMessage = (type, data = {}) =>
...
-
 const composeEnhancers = composeWithDevTools({});
 
 const store = createAppStore(
   {},
   composeEnhancers(
     applyMiddleware(
-      updateExtensionThemeMiddleware,
-      updateHistoryMiddleware,
+      ...setupMiddleware({ postMessage, urlEncodeTheme, storage }),
```
lmorchard: Bit of refactoring here, moved all the middleware into a separate module and added the stuff for image management between web app and add-on

src/web/lib/components/ThemeBuilder/index.js
```diff
@@ -2,37 +2,34 @@ import React from "react";
...
-        <Banner {...props} />
-      )}
+      {!hasExtension && <Banner {...props} />}
       {hasExtension && (
-        <div className="theme-share-save">
-          <ThemeUrl {...props} />
-          <ThemeSaveButton {...props} />
-          <GeneratorButtons {...{ theme, setTheme }} />
-        </div>
+        <React.Fragment>
```
lmorchard: This fragment should probably go away as part of visual polish, when we find a better place for the background management stuff.

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -0,0 +1,367 @@
...
+
+class Modal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.el = document.createElement("div");
+  }
+
+  componentDidMount() {
+    modalRoot.appendChild(this.el);
+    setTimeout(() => this.el.classList.add("fade-in"), 10);
```
lmorchard: This feels kind of dirty, but couldn't think of a better way to give these dialogs a transition in. Could probably also use a timed transition out, or maybe self-dismiss on any click

src/web/lib/middleware.js
```diff
@@ -0,0 +1,104 @@
...
+    },
+    [actions.ui.setSavedThemes]: (getState, dispatch, next, action) => {
+      // Watch for changes in saved themes to purge orphaned images
+      const rv = next(action);
+      purgeImages(getState, dispatch);
+      return rv;
+    }
+  };
+
+  const purgeImages = (getState, dispatch) => {
```
lmorchard: Basic garbage collection for unused images. Feels kind of dirty, but couldn't think of a better way to clean up since images are shared and a deletion in one theme doesn't mean it's not used in another.

src/web/lib/storage.js
```diff
@@ -4,76 +4,92 @@ import { normalizeTheme, themesEqual } from "../../lib/themes";
...
-    localStorage.setItem(
-      storageKey,
-      JSON.stringify({
-        theme,
-        modified: Date.now()
-      })
-    );
-    currentSavedTheme = theme;
-    notifySelfForStorage(storageKey);
+class Storage {
```
lmorchard: Refactoring here to make the localStorage handling generally usable between themes and images. Could have found a library to use here, but meh.

src/web/lib/storage.js
```diff
@@ -85,16 +101,35 @@ function notifySelfForStorage(storageKey) {
...
+    store.dispatch(actions.images.loadImage(image));
+  };
+
+  loadAllImagesIntoStore();
   updateSavedThemesInStore();
 
   window.addEventListener("storage", e => {
     log("storage event", e);
-    if (isThemeStorageKey(e.key)) {
+    if (imageStorage.isKey(e.key) && e.newValue) {
```
lmorchard: Trying to be careful here to ensure other tabs besides the current tab get image updates loaded into their redux stores

meandavejustice:
Here are some notes from my first run at reviewing the behavior and appearance.

- image shouldn't be draggable if only one item in list
- back button should remove image item, now it solely removes the thumbnail, and preview
color-no-background-image-button

- remove item, then back button, brings back item in list, without thumbnail and preview
- ui isn't under the "images and textures" tab like in the mockup, in fact there doesn't seem to be any of the tabs from the mockup
- sometimes seeing install button flash of content on scroll
color-install-addon-button-on-scroll
- image is too large message is an li
- image is too large message should be able to be dismissed.
color-oversized-image-error-interferes-with-messaging


Beyond these ui nits and things it seems to be working really well. This was a big chunk of work, I'm impressed, and can't wait to start using this feature.

lmorchard:
> Not sure how close to the linked UI this pr is aiming to be before merging

The goal in this PR is to get just the custom background part functional and close to the UX spec and then do some visual polish after merging.

> ui isn't under the "images and textures" tab like in the mockup, in fact there doesn't seem to be any of the tabs from the mockup

Yeah, we're not doing any of that in this PR.

> image is too large message is an li
> image is too large message should be able to be dismissed.

That's the kind of stuff we want to do in follow up work for visual polish. I'd say the functionality in this PR is the main thing.

> sometimes seeing install button flash of content on scroll

This might be a performance issue: if the add-on is lagging and the site doesn't get a ping response after 3 seconds, the site reverts back to the not-installed state

lmorchard:
Other than the UX polish stuff, looks like these are the functional issues so far?

* image shouldn't be draggable if only one item in list
* back button should remove image item, now it solely removes the thumbnail, and preview
color-no-background-image-button
* remove item, then back button, brings back item in list, without thumbnail and preview
* sometimes seeing install button flash of content on scroll

lmorchard:
Rebased against master

lmorchard:
IMO, if there are mainly just minor issues and this thing is mostly functional without big showstoppers -,we should merge to unblock further iteration on the feature

lmorchard:
Updated to try disabling drag with 1 item. Also bumped up the add-on uninstall detection threshold, which should affect the UI reverting back to uninstalled state. The back button stuff will take a lot more work to address, so I'd say let's punt that to a follow up issue

meandavejustice:
@lmorchard Looks like there is another conflict with `package-lock.json`. I'm happy to merge once that is updated.

opened #413 and #414 as follow up issues

PR #393, opened 2018-08-14
Upgrade eslint-plugin-import to 2.13.0

lmorchard:
I think this should fix some linting errors - extracting this from PR #391


lmorchard:
(will probably just self-merge, assuming it actually passes circleci)

PR #418, opened 2018-09-06
Retain images used in the undo/redo stack

### LINKED ISSUE #414: remove item, then back button, brings back partial item in list
item is shown without thumbnail and preview

https://github.com/mozilla/FirefoxColor/pull/391#issuecomment-416980894

lmorchard:
- Also only purge images when an explicit delete action is taken,
  since it seems like purging is too over-zealous in general

Fixes #414


meandavejustice:
Works great!

PR #421, opened 2018-09-07
Fix ThemeUrl updating and drop deprecated componentWillReceiveProps

lmorchard:
Just noticed that copying from the share field in the page was broken - probably from the custom background PR. This should fix and eliminate a deprecated API warning


lmorchard:
Doing a quick self-merge

PR #445, opened 2018-09-21
Clean up image ui

johngruen:
I think this will actually fix most of the bugs in the handoff milestone


src/web/lib/components/ThemeBackgroundPicker/index.js
```diff
@@ -1,112 +1,16 @@
 import React from "react";
```
johngruen: I just renamed some files here...what used to be the ThemeBackgroundPicker is not the ThemePatternPicker

meandavejustice:
Overall works/looks good. Found a couple issues:

1. image thumbnails can overlap depending on the dimension/orientation of image
2. When uploading there is a long delay (10 seconds) before the thumb and background updates with the new image.
3. The `Add Another` button disappears when 3 images are uploaded

all should be shown in this image:
color-image-delay

johngruen:
@meandavejustice updated PR does the following:


> Overall works/looks good. Found a couple issues:
> 
>     1. image thumbnails can overlap depending on the dimension/orientation of image
fixed in amended commit
> 
>     2. When uploading there is a long delay (10 seconds) before the thumb and background updates with the new image.
I'm not seeing this, and i don't think this behavior would have been introduced in this PR. I changed the image loading behavior to make it more obvious, but maybe file a separate bug to track, since this PR is largely cosmetic
> 
>     3. The `Add Another` button disappears when 3 images are uploaded
> 
This is intended
> 
> all should be shown in this image:
> color-image-delay

PR #486, opened 2018-10-03
Improve accessibility

sorin-davidoi:
[aXe](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/) can be used to observe the fixed violations.

### DIFF
--- a/src/web/lib/components/ThemeBuilder/index.js
+++ b/src/web/lib/components/ThemeBuilder/index.js
@@ -104,28 +104,66 @@ export const ThemeBuilder = props => {
     }
   };
 
+  let tabsElement = null;
+
+  const handleKeyUp = e => {
+    const currentTab = document.activeElement;
+    const currentTabIndex = Array.from(tabsElement.children).indexOf(currentTab);
+
+    let nextTab = null;
+
+    switch (e.key) {
+      case "ArrowLeft": {
+        nextTab = tabsElement.children[Math.max(0, currentTabIndex - 1)];
+        break;
+      }
+      case "ArrowRight": {
+        nextTab = tabsElement.children[Math.min(tabList.length - 1, currentTabIndex + 1)];
+        break;
+      }
+      case "Home": {
+        nextTab = tabsElement.children[0];
+        break;
+      }
+      case "End": {
+        nextTab = tabsElement.children[tabList.length - 1];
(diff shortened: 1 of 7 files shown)


src/web/lib/components/ThemePatternPicker/index.js
```diff
@@ -1,27 +1,38 @@
 import React from "react";
-import classnames from "classnames";
 
 import "./index.scss";
 
 import { colorToCSS } from "../../../../lib/themes";
 import { bgImages } from "../../../../lib/assets";
 import Metrics from "../../../../lib/metrics";
 
+console.log(bgImages);
```
lmorchard: Looks like a debugging leftover that should get deleted?

src/web/lib/components/ThemeBuilder/index.js
```diff
@@ -104,28 +104,66 @@ export const ThemeBuilder = props => {
...
+      nextTab.setAttribute("tabIndex", 0);
+      nextTab.focus();
+    }
+  };
+
   return (
     <div className="theme-builder">
       <div className="theme-builder__tabs-wrapper">
-        <ul className="theme-builder__tabs">
+        <div className="theme-builder__tabs" role="tablist" ref={el => (tabsElement = el)} onKeyUp={handleKeyUp}>
```
lmorchard: Nit: There's [a newer ref API in React 16.3](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) that uses an object created in the constructor rather than a function created at render time. Might be nice to use that here. 

But, of course, we're still using this function form in other places, so 🤷‍♂️

src/web/lib/components/ThemePatternPicker/index.js
sorin-davidoi: Indeed, fixed.

src/web/lib/components/ThemeBuilder/index.js
sorin-davidoi: The documentation states

> You may not use the ref attribute on functional components because they don’t have instances.

lmorchard: Ah, good call, I missed that!

lmorchard:
Thanks for the patch!

PR #494, opened 2018-10-05
Export static theme add-on .xpi from current theme settings

lmorchard:
- Enable Export button in AppHeader

- New export.js module with client-side support to assemble a .zip/.xpi
  static theme add-on

- Add stub component for ExportThemeDialog

- Add Redux store support for theme export progress and result

- Extract shared convertToBrowserTheme code from add-on for producing
  real theme from Redux data

- Add jszip and redux-promise dependencies

- Turn off some eslint rules that seem annoying

Issue #93


lmorchard:
To use this, you should be able to hit "Export" in the header and then see links to theme.zip & theme.xpi in the browser preview area. They're the same file, but I found .zip easier to download and open to look at than .xpi. But, you can use the .xpi in about:debugging to load up the theme

capture

lmorchard:
Oh yeah, and we could do with adding some way to let the user set the name, version, etc for the add-on on export.

src/web/lib/export.js
```diff
@@ -0,0 +1,102 @@
+import JSZip from "jszip";
```
sorin-davidoi: This is a pretty big package (https://bundlephobia.com/result?p=jszip@3.1.5) that will be used by a small fraction of the users loading the website. Could we try dynamically importing it when `performThemeExport` is called?

lmorchard: I forgot that Webpack makes this pretty easy, so [this little tweak should do import `performThemeExport` and dependencies on demand](https://github.com/mozilla/FirefoxColor/pull/494/files#diff-462b9ae37a974155d575e827e27c7076R144)

meandavejustice:
This works for me, really cool feature.

PR #516, opened 2018-10-12
Improved UI for export completion

lmorchard:
- Extract common dialog styles into mixins

- Move export complete dialog into top of page

- Rework export dialog

- Disable export button during export


lmorchard:
We probably want to expand this with some copy explaining what to do with the .zip or .xpi after downloading. Maybe only offer the .xpi also. 🤷‍♂️

meandavejustice:
Works for me locally, :+1:

lmorchard:
Looks like PR #524 supersedes this one, so I'll just close for now

PR #524, opened 2018-10-17
export dialog polish

johngruen:
@lmorchard this branch makes style changes to the export branch... I added a step to add a title to the theme but didn't wire up any of the redux to do so. I suppose we'd also need some minimal error handling, or at least to not let users advance to the download step until they'd input a name.


src/web/lib/components/ExportThemeDialog/index.js
```diff
@@ -1,27 +1,113 @@
...
+              Next
+            </button>
+          </>
+        );
+      case 1:
+        return (
+          <>
+            <h2>Name your theme</h2>
+            <form className="modal__form">
+              <label>Give your theme a descriptive name.</label>
```
sorin-davidoi: The label should be associated with the input.

src/web/lib/components/ExportThemeDialog/index.js
```diff
@@ -1,27 +1,113 @@
...
+              <input type="text" maxLength="32" placeholder="Theme name" />
+            </form>
+            <button className="modal__button" onClick={() => this.handleNext()}>
```
sorin-davidoi: Having the button inside the form with `type="submit"` would allow to proceed to the next modal by pressing `Enter` while the focus is on the input field.

lmorchard:
@johngruen Heya - I made some further tweaks to wire up the theme name field. You might want to check it out. All the rest looks okay by me!

PR #637, opened 2018-12-17
Remove feedback button

### LINKED ISSUE #626: Remove feedback link and references to Test Pilot from app
For graduation. This can be done whenever.

meandavejustice:
- fixes #626


src/web/lib/components/AppHeader/index.js
```diff
@@ -7,7 +7,6 @@ import {
 } from "../../../../lib/generators";
 
 import { version } from "../../../../../package.json";
```
lmorchard: Looks like this needs to be snipped out too, since the survey link on line 39 was its only use.

PR #795, opened 2019-04-30
pushing-commits-to-development-branch

mgillespy:
Pushing commits from October 2018 - April 2019 to development branch.


lmorchard:
FWIW, for future pushes to `development` / `stage` / `production`, I wouldn't even bother with a pull request. You can just force-push the top of master to development from the command line, something like so:

```
git co development
git reset --hard master
git push -f origin development
```

You can do the same for `stage` and `production` branches instead of `development`, too. And if you want to better control what's released, you can use git tags or github releases and `git reset` using one of those. But, we never got that precise on releases in the Test Pilot period.

PR #803, opened 2019-05-23
Fixes 763: update properties deprecated in Firefox 69

mgillespy:

PR #819, opened 2019-08-07
fix: update theme properties to remove usage of the deprecated lwt aliases

### LINKED ISSUE #763: Update properties that will be deprecated in Firefox 69
In [Firefox 69](https://wiki.mozilla.org/Release_Management/Calendar), the following theme properties will be deprecated: 

- toolbar_text
- accentcolor
-textcolor

[MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme#Aliases) recommends replacing themes with the following: 

-bookmark_text
-frame
-frame_inactive
-tab_background_text

We need to replace the deprecated properties with the recommended replacements prior to Firefox 69 going into the Beta channel (July 9, 2019).

rpl:
This PR includes and supersedes #803 
Besides the change already included in #803, this PR also brings back toolbar_text (as it is not on the deprecation path anymore) and remove the changes to the package.json file (because they are unrelated to this fix and they can be applied in separate PRs).

NOTE: the theme normalization happens on the companion website and so to fully fix the incompatibility between Firefox Color and Firefox 70 we need to both publish the new version of the xpi and update the color.firefox.com website.

Fixes #763

### DIFF
--- a/docs/theme-schema.json
+++ b/docs/theme-schema.json
@@ -79,7 +79,7 @@
                 "toolbar_field_text": {
                     "$ref": "#/definitions/rgb"
                 },
-                "bookmark_text": {
+                "toolbar_text": {
                     "$ref": "#/definitions/rgb"
                 },
                 "tab_line": {
@@ -92,7 +92,7 @@
                     "$ref": "#/definitions/rgb"
                 }
             },
-            "required": ["toolbar", "bookmark_text", "frame", "tab_background_text", "toolbar_field", "toolbar_field_text", "tab_line", "popup", "popup_text"],
+            "required": ["toolbar", "toolbar_text", "frame", "tab_background_text", "toolbar_field", "toolbar_field_text", "tab_line", "popup", "popup_text"],
             "additionalProperties": false
         },
         "images": {
--- a/docs/theme-schema.json
+++ b/docs/theme-schema.json
@@ -64,10 +64,10 @@
         "colors": {
             "type": "object",
             "properties": {
-                "accentcolor": {
+                "frame": {
                     "$ref": "#/definitions/rgb"
                 },
-                "textcolor": {
+                "tab_background_text": {
                     "$ref": "#/definitions/rgb"
                 },
                 "toolbar": {
@@ -79,7 +79,7 @@
                 "toolbar_field_text": {
                     "$ref": "#/definitions/rgb"
                 },
-                "toolbar_text": {
+                "bookmark_text": {
                     "$ref": "#/definitions/rgb"
                 },
                 "tab_line": {
@@ -92,13 +92,13 @@
                     "$ref": "#/definitions/rgb"
                 }
             },
-            "required": ["toolbar", "toolbar_text", "accentcolor", "textcolor", "toolbar_field", "toolbar_field_text", "tab_line", "popup", "popup_text"],
+            "required": ["toolbar", "bookmark_text", "frame", "tab_background_text", "toolbar_field", "toolbar_field_text", "tab_line", "popup", "popup_text"],
             "additionalProperties": false
         },
         "images": {
             "type": "object",
             "properties": {
-                "headerURL": {
+                "theme_frame": {
                     "type": "string"
                 },
                 "additional_backgrounds": {
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -23,9 +23,9 @@ export const CUSTOM_BACKGROUND_DEFAULT_ALIGNMENT = "right top";
 
 export const colorLabels = {
   toolbar: "Toolbar Color",
-  toolbar_text: "Toolbar Icons and Text",
-  accentcolor: "Background Color",
-  textcolor: "Background Tab Text Color",
+  bookmark_text: "Toolbar Icons and Text",
+  frame: "Background Color",
+  tab_background_text: "Background Tab Text Color",
   toolbar_field: "Search Bar Color",
   toolbar_field_text: "Search Text",
   tab_line: "Tab Highlight Color",
@@ -35,7 +35,7 @@ export const colorLabels = {
 
 export const fallbackColors = {
   popup: "toolbar",
-  popup_text: "textcolor"
+  popup_text: "tab_background_text"
 };
 
 export const colorsWithAlpha = ["toolbar", "toolbar_field"];
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -23,7 +23,7 @@ export const CUSTOM_BACKGROUND_DEFAULT_ALIGNMENT = "right top";
 
 export const colorLabels = {
   toolbar: "Toolbar Color",
-  bookmark_text: "Toolbar Icons and Text",
+  toolbar_text: "Toolbar Icons and Text",
   frame: "Background Color",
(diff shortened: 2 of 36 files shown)


rpl:
@ntim would you mind to double-check this PR? it is a rebased (and sightly updated) version of #803

rpl:
@robwu noticed some additional issues with the original version of these changes:
- by calling normalizeThemeColor before checking for the deprecated lwt aliases, the deprecated aliases are removed and the new canonical properties set to their default values
- the deprecated lwt aliases will still be used by a theme stored in the storage.local backend 

I pushed an additional patch in my branch (https://github.com/rpl/FirefoxColor/commit/816aef9469b4248a4ad44e1af1daa2ce6135b9e4), I'm not sure why github is not yet listing it as part of this PR.

To be fair we could likely move the code that looks for the deprecated properties into its own method and only call it from the `convertToBrowserTheme` method, instead of calling `normalizeTheme` again from inside the `convertToBrowserTheme`.

@robwu would you mind to take a look to the linked patch and the above comment in the meantime?

nt1m:
@rpl The main reason #803 wasn't merged yet is that saved themes aren't migrated automatically to the new properties.

Another thing that would be nice to have working is that theme links copied from the old Firefox Color (with the old props) should work in the new Firefox color.

I gave @caitmuenster a full list of the things to test last week. Caitlin, would you mind writing them down here so Luca can take a look? I unfortunately can't remember them all.

Thanks :)

Rob--W:
The patch looks good with the additional commit.

Just one thing: In case the new property is present, ignore the old property. So change

```
if (colors.accentcolor) {
```

to

```
if (colors.accentcolor && !colors.frame) {
```

Rob--W:
@nt1m I also spotted those issues when I reviewed the original PR, but @rpl has now addressed those with the third commit (that somehow does not automatically appear in this PR).

With that commit, (saved) themes with old properties are still applied as expected.

caitmuenster:
Here's the testing run down: 

- Sharing theme
- Exporting theme
- Adding custom background images 
- Migration code (will themes made using the deprecated properties on Color be migrated to use the correct properties when Color is updated)

rpl:
> Here's the testing run down:
> 
>     * Sharing theme
> 
>     * Exporting theme
> 
>     * Adding custom background images
> 
>     * Migration code (will themes made using the deprecated properties on Color be migrated to use the correct properties when Color is updated)

I briefly tried these features and they all seem to be working as expected (e.g. I manually decoded the theme shared in a theme url and it did contain only the canonical theme properties as expected, the zip generated zip theme file was also looking good)

The migration code doesn't currently automatically save the "rewritten" theme properties, because the theme is being rewritten right before being applied, @ntim could that be covered by a follow up?

@ntim it would be also great if you could provide some STR that the QA may be able to use to verify the bug manually (once this has been published in a new xpi version and the updated website deployed).

@ntim @caitmuenster let me know if there is anything else that is blocking this PR.

caitmuenster:
Thank you @rpl! 

> The migration code doesn't currently automatically save the "rewritten" theme properties, because the theme is being rewritten right before being applied, @ntim could that be covered by a follow up?

I'm ok with this! 

@Rob--W, if you could confirm this patch looked ok to you I'm happy to merge it in.

Rob--W:
- New website + new addon = OK
- New website + old addon = broken (addon does not recognize newest theme properties, and fails to fully apply the theme)
- Old website + new addon = broken (website does not recognize newest theme properties, and fails to preview the theme)
- Old website + old addon = OK.

The website does currently not support older versions of the add-on (=before this PR). Consequently, if the user edits a theme on the website and attempts to share it, then users of older Firefox Color versions will see a partial theme, because the (old) add-on does not recognize the new format. When the user updates to the new version, the theme works though.

Example: all colors should be red, "Background Color" should be green.

- New theme (with this PR): https://color.firefox.com/?theme=XQAAAAITAQAAAAAAAABBqYhm849SCia2CaaEGccwS-xNKliFvrJIaTya9DDK7s1WCD-NxveHoSVR5Zp4bI43EigXLfLtYWw5VGRrW_-bf00bwSHPBCVuZPd6wFOEumb4qa5IO0FtcZQHPorly463gJr-OvqaEz3byyqCD7e6fts2A5u6bp_bYQVeWP9xHw-w2mgjJwzd72ypUTEs9-rUxDKQV8VtJ8fm6v_dkugA
- Old theme (current production): https://color.firefox.com/?theme=XQAAAAIPAQAAAAAAAABBqYhm849SCia2CaaEGccwS-xNKliFvrJIaTya9DDK7s1WCD-NxveHoSVR5ZyK1ssMuAsVfU_WM8E_WE3n9fxb0UDtnAINpOjCVEoE5JVepPN3SqLs65MqshKQ5SVQiahjGiq3O12IIuVH2kMA-mV3qy9tAGvajEoSwQqKMB0o9ho_WmjrMbW8WL9F4oJBzNqBvtV3ROwleXzMnQIT__SUngA

When I open "New theme" in the current https://color.firefox.com/ website, then "Background Color" and "Background Tab Text Color" have the unexpected color (expected green+red, got the default colors instead). If the website were to be updated, the "New theme" links are previewed as expected, but old Firefox add-ons will fail to apply the theme.

I suggest that we also include the theme with the original key in the message, so that older versions of Firefox Color will continue to recognize the theme. A quick 'n dirty way to achieve that is by appending the following logic to https://github.com/mozilla/FirefoxColor/blob/17a2fdacd04eefb45c27dc6910de0d089bcd6669/src/web/index.js#L54 :
```javascript
// Add aliases for Firefox Color 2.1.4 and earlier.
let {theme} = data;
if (theme) {
  if (!theme.accentcolor && theme.frame) {
    theme.accentcolor = theme.frame;
  }
  if (!theme.textcolor && theme.tab_background_text) {
    theme.textcolor = theme.tab_background_text;
  }
}
```

@rpl Could you add that change?

rpl:
> @rpl Could you add that change?

@Rob--W sure, I just took a look and added a change similar to the one you suggested (in 7d883c4), the main differences are:
- the theme colors are inside the theme.colors properties
- also included the theme images.headerURL alias
- restricted the behavior to postMessage calls with type "setTheme"
- expanded the inline comment a bit

Let me know how the changes applied looks to you (and if there is anything else to do that I may have missed in your last review comment).

caitmuenster:
Thanks @Rob--W and @rpl! Merging this now.

gaby2300:
@caitmuenster Does this mean the background issue in Firefox Color is fixed?

caitmuenster:
@gaby2300 We're almost there! This means the change is in the code, but the code still needs to be deployed. Timeline for that is still TBD.

gaby2300:
@caitmuenster Thanks for your answer! It this the issue I have to follow/subscribe to to know when will this be done? If not, is there another one?

caitmuenster:
@gaby2300 We'll close the original issue once this has been deployed to production. :)

marcin-kozinski:
@caitmuenster Do you mean #763? It autoclosed when this PR was merged :D

Rob--W:
FYI there are some regressions that are caused by this PR. I will submit a new PR to fix those.

PR #824, opened 2019-08-14
Add ADDON_URL for building site

jbuck:


jbuck:
> @caitmuenster: jorge says that if there is no update URL in the .xpi on the website, it will check for updates on AMO
> @caitmuenster: if there is an update URL, a couple of steps are needed to move them to AMO but it’s doable

There is an update URL in the XPI https://testpilot.firefox.com/files/FirefoxColor@mozilla.com/updates.json so we may need to wait to merge/deploy this change

caitmuenster:
@Rob--wu would you mind reviewing this? It's a little outside of my area of expertise. :)

.circleci/config.yml
```diff
@@ -5,6 +5,8 @@ jobs:
   build_and_test:
     docker:
       - image: circleci/node:8.11.2@sha256:495a13db597b5306f05ae728489fedb9faa4defbac24ca0e00e8cc1eca95396d
+    environment:
+      ADDON_URL: https://addons.mozilla.org/firefox/downloads/latest/firefox-color/
```
Rob--W: Although it doesn't really matter for the validity of the URL, let's add "firefox-color.xpi" so that the link ends with a ".xpi" extension.

PR #825, opened 2019-08-16
fix: Fix theme migration logic, follow-up to #819

Rob--W:
See commit message.

This is a follow-up to #819.

### DIFF
--- a/src/lib/themes-test.js
+++ b/src/lib/themes-test.js
@@ -1,10 +1,11 @@
 const { expect } = require("chai");
-const { presetThemesContext } = require("./assets");
+const { presetThemesContext, bgImages } = require("./assets");
 const themes = require("./themes");
 
 describe("lib/themes", () => {
   describe("normalizeThemeColor", () => {
     const subject = themes.normalizeThemeColor;
+
     it("should properly handle alpha channel in colors with alpha", () => {
       const input = { r: 255, g: 255, b: 255, a: 0.5299999713897705 };
       const expected = { r: 255, g: 255, b: 255, a: 0.53 };
@@ -22,6 +23,7 @@ describe("lib/themes", () => {
 
   describe("normalizeTheme", () => {
     const subject = themes.normalizeTheme;
+
     it("should not change preset themes", () => {
       presetThemesContext.keys().forEach(name => {
         // There are some acceptable differences from preset themes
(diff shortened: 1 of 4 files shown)


src/lib/themes-test.js
```diff
@@ -33,6 +33,62 @@ describe("lib/themes", () => {
         expect(output).to.deep.equal(theme);
       });
     });
+    it("should migrate themes without popup/popup_text", () => {
+      const theme = {
+        images: {
+          headerURL: ["dummy.png"],
```
rpl: shouldn't images.headerURL also be in the expected theme along with "images.theme_frame"?

Rob--W: Yes. Apparently I need one of the files in `src/images/patterns` for the test to pick it up. I've updated the test.

Rob--W:
> at a first glance it seems to ne that the "legacy LWT aliases" theme properties are now going to still be part of the theme also in the new version of the extension, am I right?
>
> If that is the case, it may be reasonable to still remove them on the extension side to avoid logging warnings about them in Firefox >= 70 (mostly because they may be misleading).

This is not the case. `convertToBrowserTheme` is used to create the theme for use in the `browser.theme` API. These use the return value of `normalizeTheme`, and are independent of the original input parameter.

I've added a unit test (the first one!) that verifies that `convertToBrowserTheme` does what it's supposed to be doing.

src/lib/themes-test.js
```diff
@@ -33,6 +33,137 @@ describe("lib/themes", () => {
...
+        },
+        title: ["old theme without popup"],
+      };
+      const output = subject(theme);
+      expect(output).to.deep.equal(expectedTheme);
+    });
+    it("should migrate deprecated theme properties", () => {
+      const deprecatedTheme = {
+        images: {
+          headerURL: ["dummy.png"],
```
rpl: headerURL is meant to be a string

also, now I see that only background images that are included in bgImages are meant to be still part of the normalized theme, would you mind to add an inline comment on top of this line to mention that it is expected that dummy.png will not be part of `images.additional_backgrounds` in the normalized theme?

src/lib/themes-test.js
```diff
@@ -33,6 +33,146 @@ describe("lib/themes", () => {
         expect(output).to.deep.equal(theme);
       });
     });
+    it("should migrate themes without popup/popup_text", () => {
```
rpl: Nit, if you end up to force push / update this PR again, let's add an empty line between the `it(...)` calls (don't bother if you are not going to update it again for now).

Rob--W: Added a few blank lines for consistency.

caitmuenster:
Testing prod extension on https://mozilla.github.io/: 

Color-Properties-Fixed-Dev

PR #839, opened 2019-09-09
I586 displaying local storage info

rebmullin:
Here are some screenshots to show the different views:

# low storage used:
 Alt text

# over 9MB:
Alt text

# will go over the allowed storage and so doesn't save (this displays for 5 seconds):
Alt text

Also, are you okay with the showing of 3 decimal places? If the user is only adding themes (w/o images), I believe it will be below 1MB for a while so I added that to show it was "moving". 

note: I will fix the linting issues tomorrow:)


src/web/lib/components/StorageSpaceInformation/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+export const STORAGE_WARNING_MESSAGE =
+  "You have almost reached your storage limit! You may need to delete some saved images/themes to make space.";
+
+export const STORAGE_ERROR_MESSAGE_DURATION = 5000;
+
+export const localStorageSpace = () => {
+  let data = "";
+  for (let key in window.localStorage) {
+    if (window.localStorage.hasOwnProperty(key)) {
```
Rob--W: The size of the keys should also be included in the count.
```
let size = 0;
for (let [key, value] of Object.entries(localStorage)) {
  size += key.length + value.length;
}
```

The limit can be configured through the `dom.storage.default_quota` pref, defaulting to 5120 kB (= `5 x 1024`), i.e. 5242880 "bytes" (the actual implementation doesn't count the number of bytes, but the length of the string, which could be way more if the string contains multi-byte characters).

This means that the limit below (10 MB) overestimates the amount of available storage. The actual limit is `5 * 1024 * 1024`.

Firefox Color doesn't efficiently save images either: it base64-encodes images, which increases the storage size of images by 33%. This means that the total size of images is at most 5  / 1.33 = 3.76 MB. Minus other metadata (themes and such), there is not that much storage left :/

rebmullin: thank for the feedback @Rob--W :) I can adjust the calculation as suggested. 
yes, we realized the images were the main culprit (we talked about possibly compressing these but I did not include that with this PR). 

I'll take a look this evening:)

caitmuenster:
Thank you @rebmullin! Let's put some line breaks in on the text warnings. 

#### Over 9mb
You have almost reached your storage limit! (break) 
You may need to delete some images or themes to make space. 

#### Will go over the allowed storage and so doesn't save (this displays for 5 seconds):
Sorry! This cannot be added because you are over your storage limit. (break) 
Please delete some images or themes to make space. 

## Positioning
It's a little squishy on the "Custom Backgrounds" tab before any changes are made. Can we add another line or two of space? 

Screen Shot 2019-09-09 at 4 03 18 PM

## Icon
We might need an icon to indicate that the text is related to storage space. Dev tools has this one: 

image (2)

We would probably need to cut the "Storage" text off and make sure it's the right size/in alignment with the test. I don't have specific specs but I could probably create something tomorrow. @rebmullin  & @Rob--W, any thoughts?

Rob--W:
The devtools icon is a vector image, available at: https://searchfox.org/mozilla-central/rev/250f5cc9fb8bdcbb6b23d2a06acfd48addb2f99b/devtools/client/themes/images/tool-storage.svg

Direct link: https://hg.mozilla.org/mozilla-central/raw-file/d3741459641932fa5735692e20c6464ea2597c59/devtools/client/themes/images/tool-storage.svg

Before merging, could you check whether themes can still be created and exported even if the storage is full?

src/web/lib/components/StorageSpaceInformation/index.js
rebmullin: ok so I've modified the logic to include the key length as well. I have also adjusted this to be `[number]MB out of 5.242MB` (instead "..out of 10MB"):

``` let size = 0;
  for (let [key, value] of Object.entries(localStorage)) {
    size += key.length + value.length;
  }

  const sizeMB = size / 1000000;
  return parseFloat(sizeMB.toFixed(3));
```

Originally I was hitting the ~10MB when I was nearing the storage limit but w/ the updated calculation it's about this 5.24MB. Please let me know if I have this right. If so I'll push this up:)

rebmullin:
@Rob--W,

thanks so much for the icon links!

>Before merging, could you check whether themes can still be created and exported even if the storage is full?

Is the best way to check this is to unhide the "Export" button locally? Because I can't actually install the extension via localhost, this doesn't show up unless I remove the logic around it. Would that be the only way to test this?

src/web/lib/components/StorageSpaceInformation/index.js
Rob--W: Your chosen limit looks right. Do note that the limit is dependent on the browser settings, and that most users will have the default of 5 MiB. The limit should be informative, but not prevent the user from adding more images (and updating and exporting themes) if they can (to test that, visit `about:config` and set `dom.storage.default_quota` to a higher value, or simply change your application logic to "enforce" a lower-than-maximum limit). Similarly, if the actual limit is lower, then the application should still handle that gracefully.

Rob--W:
For local development, you should be able to run the extension on localhost with `npm run-script run`, as defined at:

https://github.com/mozilla/FirefoxColor/blob/444d50d1dc12fb20b52029000bdb9cb74148692c/package.json#L48

src/web/lib/components/StorageSpaceInformation/index.js
rebmullin: @caitmuenster, I know we talked about this but it does look like the default limit is ~5.24MB after adjusting the calculation as noted above. So I've updated the message to look something like the following ~ :



Also, as far as I know there is no way to find out if the user has changed the default so I am not sure if we want to make an additional note about this or this will be okay as is given most users will probably have the default. 

@Rob--W,
Maybe I am misunderstanding what you mean here:

>The limit should be informative, but not prevent the user from adding more images..

If the storage is full, it won't allow additional images to be added so in this case it would prevent the user from adding more images unless these are cleared out.

Regardless, the user will see a message ~ "Sorry! This cannot be added because you are over your storage limit...." if it will put them over but again the "..out of [storageLimit]" message <= might not be completely accurate if they're not using the default. 

Also, I noticed if the image is not in local storage (it can be removed by the X), it won't be included in the export (cc @caitmuenster). They would need to export it before removing it if they want to include it.

rebmullin:
@Rob--W 
> Before merging, could you check whether themes can still be created and exported even if the storage is full?

Looking at the zip, the export looks good to me but again if you remove the image, this will remove it from local storage so it would not be included in the export. So as long as that is there, it should be okay.

As far as creating new themes you cannot create new themes if local storage is full.

rebmullin:
hey @Rob--W, I think this should be ready now. Can you please take another look when you can :)

caitmuenster:
Hey @rpl, since @Rob--W is at Rust training for the next few days, would you be able to review this PR?

src/web/lib/components/PaginatedThemeSelector/index.js
```diff
@@ -27,71 +28,81 @@ export const PaginatedThemeSelector = ({
...
-        <footer className="page-selector">
-          <button
-            className={classNames("previous", { disabled: prevDisabled })}
-            disabled={prevDisabled}
-            onClick={
-              prevDisabled ? () => {} : () => setCurrentPage(currentPage - 1)
+    <>
+      <div className={classNames("theme-selector", className)}>
+        {page.map(([key, { theme }]) => {
+          const customImages = (theme.images.custom_backgrounds || []).map(
```
rpl: This looks almost the same thing done in `src/web/lib/components/App/index.js` at line 103, but in that version we are returning an empty object if we are unable to find the custom image, while here we are returning `{ ...item }` (as in the previous version of the code in `src/web/lib/components/App/index.js`).

it there a reason for the change applied to the function defined in `src/web/lib/components/App/index.js` that is not needed here?

caitmuenster:
> have we also evaluated to move from localStorage to IndexedDB?

I talked to @rpl about this earlier today -- I think we need to defer this piece to later in the project because we have limited time under contract and a few more p1s to work through. One of the things we want to implement to mitigate the storage issue is ask users to submit their themes to AMO so they'll have a saved copy and will be able to delete images/themes from the Fx Color UI. 

If we still have hours remaining after we work through the next couple of p1s, we can definitely revisit migrating to IndexedDB. :) 

Since this issue isn't a blocker, we'll wait for @Rob--W to finish his training and review the patch. :)

src/web/lib/components/PaginatedThemeSelector/index.js
```diff
@@ -27,71 +28,81 @@ export const PaginatedThemeSelector = ({
...
-      {pageCount > 1 && (
-        <footer className="page-selector">
-          <button
-            className={classNames("previous", { disabled: prevDisabled })}
-            disabled={prevDisabled}
-            onClick={
-              prevDisabled ? () => {} : () => setCurrentPage(currentPage - 1)
+    <>
+      <div className={classNames("theme-selector", className)}>
+        {page.map(([key, { theme }]) => {
```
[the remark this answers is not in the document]
rebmullin: hey @rpl, thanks for the review:) yes, I think that looks better/easier to read. I'll clean it up and update the PR.

src/web/lib/components/PaginatedThemeSelector/index.js
rebmullin: yes, you're right. good catch. The App file didn't have this extra check here (on line 37) so it didn't cause errors like the one in App. So yes, I'll either refactor this into a function or I'll update the logic in the App - so it's also returning an object with the undefined image property instead of the empty object. It seems either is okay but I'll keep it consistent with this one.

thanks:)

src/web/lib/components/ThemeSaveButton/index.js
```diff
@@ -8,15 +13,29 @@ export const ThemeSaveButton = ({
...
+    try {
+      let space = localStorageSpace();
+      themeStorage.put(themeStorage.generateKey(), theme);
+      setThemeBuilderPanel(3);
+      setUsedStorage({ space });
+    } catch (err) {
+      console.error(err);
+      setStorageErrorMessage(STORAGE_ERROR_MESSAGE);
+      setTimeout(
+        () => setStorageErrorMessage(""),
```
Rob--W: Could you move the logic that clears the error to the component that is responsible for rendering the error? And consider clearing the error message if the images were successfully saved.

In the current patch, anything that modifies storage is expected to call `setUsedStorage` afterwards, and apparently also update the UI if needed. Would it be feasible to trigger the UI update from the storage module itself instead, and have consistent handling (e.g. disappearing error message after some timeout)?

There is currently logic to synchronize changes to the storage from other pages (e.g. if the Firefox Color website is opened in two browser windows). You could also consider triggering the notification from that event: https://github.com/mozilla/FirefoxColor/blob/444d50d1dc12fb20b52029000bdb9cb74148692c/src/web/lib/storage.js#L127
(note: this event is only fired for storage changes from other pages. To detect changes by the current page, you need to update the `.put` / `.delete` methods.)

src/web/lib/components/StorageSpaceInformation/index.js
```diff
@@ -0,0 +1,61 @@
...
+            You have almost reached your storage limit! <br /> You may need to
+            delete some images or themes to make space.
+          </span>
+        </div>
+      )}
+      {props.storageErrorMessage && (
+        <strong className="storage-space-information-error">
+          <div
+            className="storage-space-information-warning"
+            dangerouslySetInnerHTML={{
```
Rob--W: Avoid `dangerouslySetInnerHTML` when possible. It seems that there are only two possible values, empty string or `STORAGE_ERROR_MESSAGE`. Neither really requires HTML, so just treat the message as plain text.

rebmullin: I had markup in the string but yes I can definitely redo and style the wrapping div to get the same effect.

src/lib/store.js
```diff
@@ -263,7 +281,21 @@ export const reducers = {
   theme: undoable(
     handleActions(
       {
-        SET_THEME: (state, { payload: { theme } }) => normalizeTheme(theme),
+        SET_THEME: (state, { payload }) => {
```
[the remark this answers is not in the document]
rebmullin: So I wrote a test but after seeing a bug :/ (described below a bit), I am removing this altogether. I think there is still UI things to be worked out here (or at least understood on my part).

src/web/lib/components/App/index.js
```diff
@@ -99,7 +102,7 @@ export const AppComponent = props => {
 
   const customImages = (theme.images.custom_backgrounds || []).map(item => {
     const customImage = { ...item };
-    customImage.image = themeCustomImages[item.name].image;
+    customImage.image = themeCustomImages[item.name] && themeCustomImages[item.name].image;
```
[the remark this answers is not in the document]
rebmullin: so this happens when the user saves a theme with a background image and then the user deletes the image - via the X button here - and does not delete the associated theme. We could delete the associated theme/s but my understanding was we wanted to leave these unless the user explicitly X’d these from the “Saved themes” tab. But this does cause some issues  - like `image.name` being void. Also, the theme will not be exported with the image that is deleted. I don’t exactly remember why we decided this was okay. I think it's because are future enhancements coming around this so for now this was okay. I believe with the util function I set it up, this shouldn't be an issue now since we are not including images that have been removed. I thought about updating the actually theme that references this but this was a quicker solution to address this.

With that being said, one thing I just noticed is if you select a theme in “Saved theme,” and go back to the “Custom backgrounds” tab, it only shows the images that are in localStorage images that are associated with the theme that was just selected in the “Saved theme” tab. Initially, I thought the idea of the “Custom background” tab was to show all the background images that have been uploaded but now seeing this I am not sure. I didn't realize you could click the associated theme and remove the image that way.  This is why I am removing the code I had added above (when the user goes into selecting a theme workflow, this doesn’t work as expected anyway). So given that and that you can access and potentially remove images via the “Saved themes”, I am hoping this is sufficient for now. If we do decide we want custom backgrounds to show all images that are in localstorage, I am hoping this can done in a follow up issue since it’s more involved than I initially thought.

src/web/lib/components/ThemeSaveButton/index.js
rebmullin: So I did move some stuff around. I moved the setLocalStorage to this eventListener function and I moved the setTimeout to the StorageSpaceInformation component (since the storage event listener doesn't get called if exceeds).  The error message should not be set if the images/themes are saved. The error message should only be set if it doesn't go through. I still have the setting part in the catch. in the catch.

I am not sure I am clear on all your suggestions here so please let me know what you think.

Rob--W: The "storage" event is only fired if the storage is modified by another page (not the current page). So with your current change, `actions.ui.setUsedStorage` seems to only be dispatched once in practice: When the `lib/storage.js` module is executed.
If you want to automatically propagate storage size changes, then you have to also dispatch `setUsedStorage` whenever the storage is modified by the `put` and `delete` methods in `lib/storage.js`.

> The error message should not be set if the images/themes are saved. The error message should only be set if it doesn't go through.

If the error condition is false (i.e. enough storage), then any previously shown error message should be removed. Hence, I expected the `setStorageErrorMessage` method to be called unconditionally, potentially with an empty string.

>  I still have the setting the error in the catch...

I was thinking of handling storage errors in the `put` method in `lib/storage.js`, so that the UI is always updated regardless of where `put` was called. But wrapping the `put` method in try-catch and calling `setStorageErrorMessage` also works, so it's fine to keep it there.

src/web/lib/components/StorageSpaceInformation/index.js
```diff
@@ -0,0 +1,67 @@
...
+  const sizeMB = size / 1000000;
+  return parseFloat(sizeMB.toFixed(3));
+};
+
+export const StorageSpaceInformationComponent = props => {
+  React.useEffect(() => {
+    if (props.storageErrorMessage.length > 0) {
+      setTimeout(() => props.setStorageErrorMessage(""),
+        STORAGE_ERROR_MESSAGE_DURATION
+      );
```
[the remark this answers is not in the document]
rebmullin: yes, I meant to add the `clearTimeout`. thanks for the reminder!

Regarding useEffect, it says the following on react's website:
>Don’t call Hooks inside loops, conditions, or nested functions. Instead, always use Hooks at the top level of your React function...

REF: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

rebmullin:
Thanks @Rob--W 

>The "storage" event is only fired if the storage is modified by another page (not the current page)...

I am not clear when you say "another page"? Do you mean another window? I'm so sorry but can you clarify how I can see this? I am not clear on how to reproduce this so I can see what the issue is exactly :/

>If the error condition is false (i.e. enough storage), then any previously shown error message should be removed. Hence, I expected the `setStorageErrorMessage` method to be called unconditionally, potentially with an empty string.

Sure, I can look into this. I thought I could add this to the same storage event listener but this doesn't seem to work when I add an image - and it goes over - so I'll have to look into this 1 a bit more it seems.

>I was thinking of handling storage errors in the put method in lib/storage.js, so that the UI is always updated regardless of where put was called. But wrapping the put method in try-catch and calling setStorageErrorMessage also works, so it's fine to keep it there.

I initially did think about moving both the setStorage functions to the put/delete methods themselves but I'd have to get access to the store so I just thought it might be easier to do it this way. I looks like I'll have to revisit this given the issue i had above with the image and your first comment though.

Rob--W:
> Thanks @Rob--W
> 
> > The "storage" event is only fired if the storage is modified by another page (not the current page)...
> 
> I am not clear when you say "another page"? Do you mean another window? I'm so sorry but can you clarify how I can see this? I am not clear on how to reproduce this so I can see what the issue is exactly :/

Self-contained test case, open example.com twice. Then open the console for the first and run:

```javascript
window.addEventListener("storage", console.log);
localStorage.x = 1;
// note that console.log was not called, i.e. storage event not fired.
```

Then open the devtools console in the second tab, and run:

```
localStorage.x = 2;
```

Switch back to the first tab, and notice that the storage event was triggered, as `console.log` has been called.

Note that the documentation also states:

> 
The storage event of the Window interface fires when a storage area (localStorage or sessionStorage) has been **modified in the context of another document**.

(emphasis mine, source: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event)

rebmullin:
thanks for the explanation @Rob--W :) The example above helps but I do not see this happening on this app. I see an implemented "hack" in our code which supposedly fixes this:

https://github.com/mozilla/FirefoxColor/blob/master/src/web/lib/storage.js#L93

Shouldn't fix this problem you describe above?

Anyway, I ended up leaving `setStorageSpace` in the storage event listener since this does help keep the space info in sync in the 2 tab scenario but I moved the try/catch to the Storage's `put`. note I did try to move `setStorageSpace` to the `put` but it wasn't keeping the storage info in sync between the 2 tabs this way. If this is an actual use case (2 be using 2 tabs), then hopefully this is okay..

PR #842, opened 2019-09-22
I841 updating addon url / install text for default/dev only-wip

rebmullin:
hey @caitmuenster, so I’ve updated the addon url in this PR. I think this might be a little difficult to test locally (since it links out) so my thinking is if we could add the FFColor addon to AMO dev (I might need access to this or help adding this addon here - to AMO dev), we could test out the flow there first. note: I think we would need to also adjust some config settings to allow the dev install there.  maybe @Rob--W can weigh in or advise on this:)

In theory though, you can also see the flow by starting at the AMO site (since this green button will link directly here) and I believe it already seems do what you are asking for here. I know we were going back and forth on this a little bit last week so let me know:)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "test:js": "mocha --require src/lib/test-setup.js --recursive \"src/**/*-test.js\"",
     "deploy": "gh-pages -x -d build/web -r \"https://$GH_TOKEN@github.com/mozilla/FirefoxColor.git\"",
     "release:base": "npm-run-all clean build:web sign && mv addon.xpi build/web && npm run deploy",
-    "release:dev": "cross-env ADDON_URL='https://mozilla.github.io/FirefoxColor/addon.xpi' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
+    "release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
   },
   "engines": {
     "node": ">=8.9.4"
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "test:js": "mocha --require src/lib/test-setup.js --recursive \"src/**/*-test.js\"",
     "deploy": "gh-pages -x -d build/web -r \"https://$GH_TOKEN@github.com/mozilla/FirefoxColor.git\"",
     "release:base": "npm-run-all clean build:web sign && mv addon.xpi build/web && npm run deploy",
-    "release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
+    "release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/firefox/addon/firefox-color/' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
   },
   "engines": {
     "node": ">=8.9.4"
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -20,7 +20,7 @@ const isDev = nodeEnv === "development";
 
 const defaultEnv = {
   NODE_ENV: nodeEnv,
-  ADDON_URL: "https://addons.mozilla.org/en-US/firefox/addon/firefox-color/",
+  ADDON_URL: "https://addons.mozilla.org/firefox/addon/firefox-color/",
   SITE_URL: siteUrl,
   DOWNLOAD_FIREFOX_UTM_SOURCE: downloadFirefoxUtmSource,
   LOADER_DELAY_PERIOD: "2000"
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -20,7 +20,7 @@ const isDev = nodeEnv === "development";
 
 const defaultEnv = {
   NODE_ENV: nodeEnv,
-  ADDON_URL: "addon.xpi",
+  ADDON_URL: "https://addons.mozilla.org/en-US/firefox/addon/firefox-color/",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
     "test:js": "mocha --require src/lib/test-setup.js --recursive \"src/**/*-test.js\"",
     "deploy": "gh-pages -x -d build/web -r \"https://$GH_TOKEN@github.com/mozilla/FirefoxColor.git\"",
     "release:base": "npm-run-all clean build:web sign && mv addon.xpi build/web && npm run deploy",
-    "release:dev": "cross-env ADDON_URL='https://mozilla.github.io/FirefoxColor/addon.xpi' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
+    "release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
```
rebmullin: hmm ..I might have to rethink this a bit if it's a problem that the lang is hardcoded here.

Rob--W: You can remove `en-US/`, and AMO will automatically redirect elsewhere.

E.g. try visiting https://addons.mozilla.org/firefox/addon/firefox-color/

Rob--W:
PS. Use `@Rob--W` instead of `@rob` to notify me.

caitmuenster:
Thanks @rebmullin! Yeah, if you would like to test this on [AMO-dev](https://addons-dev.allizom.org), you should be able to do so (although you might need to create a new account). If you run into any issues, give us a holler and we will get you set up. 

The flow looks good on my local build!

caitmuenster:
Hey @Rob--W, could you review this again? We're thinking this one might be ready to go but wanted to check in first. :)

webpack.common.js
```diff
@@ -20,7 +20,7 @@ const isDev = nodeEnv === "development";
 
 const defaultEnv = {
   NODE_ENV: nodeEnv,
-  ADDON_URL: "addon.xpi",
+  ADDON_URL: "https://addons.mozilla.org/firefox/addon/firefox-color/",
```
Rob--W: AMO provides the production version of the add-on, which doesn't work on dev sites.

You should probably keep addon.xpi here, or (preferably) at the very least add a new HTML page that in its turn links to the destination sites depending on the build type. Then developers can still find the right xpi, and it should also be obvious that the button usually points to AMO.

That HTML page doesn't have to be fancy, it can be a plain HTML page with a bunch of `<a>` on each line, maybe with a short description so that the purpose of the page and each link is obvious.

rebmullin: I thought prod yes would use this "default" url but we could override this "default" url in the dev release script here:
`"release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/firefox/addon/firefox-color/'`? 

wouldn't that override the default addon's url for the dev env? The purpose of this issue is to remove the xpi's and replace with links to AMO so the user will use the links now instead.

Also, @caitmuenster and I discussed the button text but it was decided to use `Get Firefox Color`, which is changed here:
https://github.com/mozilla/FirefoxColor/pull/842/files#diff-c24d2ffafde665f6f2fe83189e608db5R19.

let me know if I am missing something..


>The documentation should also be clarified, to explain that the normal installation flow...

I've updated the readme. Please let me know how it looks now.

Rob--W: > I thought prod yes would use this "default" url but we could override this "default" url in the dev release script here:
> `"release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/firefox/addon/firefox-color/'?`
>
> wouldn't that override the default addon's url for the dev env?

`release:dev` is only used here, when the "development" branch at Github is updated: https://github.com/mozilla/FirefoxColor/blob/444d50d1dc12fb20b52029000bdb9cb74148692c/.circleci/config.yml#L49
It is not used when you develop locally.

> The purpose of this issue is to remove the xpi's and replace with links to AMO so the user will use the links now instead.

This PR does indeed fix that.

The "Get Firefox Color" button's purpose is to enable the user to install the add-on. On the "production" branch, linking to AMO results in an add-on that works on production. On other branches, the link to AMO would not work, because the XPI from AMO is incompatible with the staging site or localhost. To avoid confusion by FirefoxColor developers, it should be clear that the button does not work.

To make sure that the local developer of FirefoxColor can find a usable XPI for testing, I previously proposed to set the default of the `ADDON_URL` to a static page that briefly explains that on production, the page would be a link to AMO, and to show links to the xpis that one can use to test locally \*. An (inferior) alternative is to document how one is supposed to install/test the add-on on development (and intentionally leave the "Get Firefox Color" button link at a target that is incompatible with the development environment).

( \* Including explaining the need to set a preference and use DevEd/Nightlly to test the install flow locally, as explained under "What are my options if I want to install unsigned extensions in Firefox?" at https://wiki.mozilla.org/Add-ons/Extension_Signing#FAQ )

rebmullin: regarding installing on dev..it should work on dev since I've added the addon to the dev AMO site (I was planning on doing the same for stage). you just need to adjust the following config settings in a profile to test this:

`extensions.webapi.testing` to `true`
`xpinstall.signatures.dev-root` to `true`

it then links out to the prod url..

Rob--W: manifest.json of the add-on at https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/ (and https://addons.mozilla.org/firefox/addon/firefox-color/) contains the following:

```
    "content_scripts": [
        {
            "js": [
                "contentScript.js"
            ],
            "matches": [
                "https://color.firefox.com/*"
            ],
            "run_at": "document_start"
        }
    ],
```

Note that the listed site is the production "https://color.firefox.com/" (based on `SITE_URL`). This add-on does not run on localhost (local development) or the stage site. Because of this, it is not possible to manually test the full installation flow.

rebmullin: ok I think I understand the problem you're explaining ... that we need to actually install the addon in its respective environment. Ok..I'll take another look this week. thanks so much:)

PR #844, opened 2019-10-07
I302 revert to default

### LINKED ISSUE #302: Allow users to revert to default themes with FF color installed
For example, I change my theme back to default and exit my browser. When I reopen my browser, it changes back to my last used Firefox Color theme

may-22-2018 14-02-20

rebmullin:
fixes #302


src/web/index.js
```diff
@@ -166,6 +166,12 @@ window.addEventListener("message", ({ source, data: message }) => {
   }
 });
 
+document.addEventListener("click", event => {
+  if (event.target.title === "Revert All") {
```
Rob--W: This is quite indirect. Why aren't you exposing a helper function through `props` via https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L200-L211 ?

src/web/lib/components/AppHeader/index.scss
```diff
@@ -57,12 +57,21 @@
     padding: 0;
     width: 64px;
 
+    span {
```
Rob--W: Undo all `pointer-events` changes here. I suppose that you did it to make sure that `event.target` pointed to the button instead of their descendant nodes, but if you add the click handler itself, then it is not needed.

As an alternative to making the child element ineligible for becoming a click target through CSS, you could also have used JS to detect whether the clicked element matches. So instead of `if (event.target.title === "Revert All")` you could have used `if (event.target.closest("[title='Revert All']")) {` (if the element is found, the matching element is returned, otherwise `null`).

src/web/lib/components/AppHeader/index.scss
```diff
@@ -57,12 +57,21 @@
...
     .app-header__button-icon {
       align-items: Center;
       display: flex;
       height: 20px;
       margin-bottom: 1px;
       width: 22px;
+      pointer-events: none;
+
+      img {
+        pointer-events: none;
```
Rob--W: This was not necessary, as `pointer-events` makes an element and all of its descendants unclickable. (The parentt elemen (`.app-header__button-icon`) was already assigned `pointer-events`).

src/extension/background.js
```diff
@@ -97,6 +97,12 @@ const messageHandlers = {
     images.forEach(name => delete customBackgrounds[name]);
     storeImages({ images: customBackgrounds });
   },
+  revertAll: (message, port) => {
+    log("revertAllThemes", message);
+    browser.theme.reset();
+    messageHandlers.setTheme(message);
```
Rob--W: `message` looks like `{type: "revertAll"}` here.

`setTheme` does `const theme = normalizeTheme(message.theme);`, and since `theme` is not set, FirefoxColor's default theme is used. This is incorrect, the purpose of the revert functionality is to use Firefox's default theme.

`setTheme` calls `storeTheme` and `applyTheme`. Since you're already calling `browser.theme.reset()`, the `applyTheme` call is unnecessary. So what remains is the need to call `storeTheme({ theme: null })`.

src/web/index.js
rebmullin: I wasn't sure about this but I wanted to use the postMessage so the background.js would be aware. I'll revisit this though..

src/extension/background.js
rebmullin: yes, I thought when this was last discussed it was decided to revert back to the default FF color theme; which is what this does.

I tried storeTheme ~:

```javascript
  browser.theme.reset();
  storeTheme({ theme: null })
```

and if you refresh the old theme is restored. let me know if I am misunderstanding your feedback though. I have to do more testing to be sure but I think the only way to keep the true default theme (the non FF color default theme), is to disable this extension or remove this default theme altogether.

Rob--W: > and if you refresh the old theme is restored.

Refresh what? If I look at src/extension/background.js, I can see that no theme is supposed to apply at extension startup. If the theme changes, it might be the result of logic in the web page.

rebmullin: > Refresh what?

if you refresh the page..

there is a default purplish theme that is applied when you first install this extension:



it has always been this way as far as I know..

Rob--W: Sorry, I should have been clearer. I wanted to ask when the theme is restored. I.e. if you click "Revert", whether that would restore the default, and stay at the default if you reload the extension **without an active Firefox Color webpage**.

When I install the Firefox Color add-on for the first time, then a new tab opens, and only then the Firefox Color **web page** changes the theme to the default.json theme.

I expect that with the patch, the Firefox Color **extension** will not change the theme after a restart. Is this the case?

Secondly, if you open the web page again, will it not touch Firefox's default theme? Based on your previous comment, it seems that the web page will change the default theme to default.json when the user visits the Firefox Color website. This should not happen.

rebmullin: > if you click "Revert", whether that would restore the default, and stay at the default if you reload the extension without an active Firefox Color webpage.

I just want to make sure we are talking about the same things, so I'm just adding some notes to clarify what is meant by "default":

- "FF Default theme" = FF's Default theme

- "*Firefox Color* Default theme" = default.json from Firefox Color (i.e. purplish/gray color that is set when you initial activate this addon).

So I tested this locally by adding the addon via `about:debugging`, clicking random and then hitting revert. Then I removed the addon via `about:debugging`. I am not 100% this simulates a live version but based on this, it does appear to keep go back to the "FF Default Theme" when I remove this extension in the way I've described above.

If I do NOT remove it, it seems to keep the "*Firefox Color* Default theme," unless the theme is changed through the extension of course. 


> I expect that with the patch, the Firefox Color extension will not change the theme after a restart. Is this the case?

If the FF color extension is active it _should_ use the last theme displayed or the "*Firefox Color* Default theme" (default here would be displayed at the beginning or after hitting "Revert.") 

If it's not active, based on my testing noted above, it _should_ be the FF theme set under customize theme.


But again I am not 100% due to testing locally - do you have a suggestion on how to test this locally to be 100% sure? :)

Rob--W: When an extension is removed, Firefox will forget about the theme sey by the (removed) extension, and consequently fall back to Firefox's default theme.

> If the FF color extension is active it should use the last theme displayed or the "Firefox Color Default theme" (default here would be displayed at the beginning or after hitting "Revert.")

When "Revert" is hit, Firefox's default theme should be used by Firefox Color.
After a browser restart, Firefox Color should use the theme before the browser/addon restart, which is Firefox's default theme.

To test the behavior of an add-on after a browser restart, you have to actually install an add-on instead of loading it as a temporary add-on via `about:debugging`/`web-ext run`.

1. Install Firefox Nightly.
2. Visit `about:config` and set `xpinstall.signatures.required` to `false`.
3. Install the xpi that is generated by `npm run xpi:dev` (this does not exist yet, but IIRC you were working on it in another bug).

In most cases, you don't need to fully restart the browser. Simply disabling then re-enabling the add-on is often a realistic test.

rebmullin: Okay so currently as I have it here, it reverts back to the default.json (the starting "canvas"). I'll have to look into this more to see if we can keep the actual FF default theme once the user hits reset..

I'll get back to you on this :)

Also, thanks for the testing notes :)

rebmullin: Hey @Rob--W 

I copied over the xpi:dev script since that PR is still pending but I have not been able to test this locally - installing the xpi and uninstalling. 

I guess I am realizing that since that PR is still pending, maybe I should have waited as it might still needs adjustments - but I wanted to verify this: that this should work locally as well?

rebmullin: hmm it seems that build/extension/manifest.json enables and disables the extension locally. I wonder if this is also a sufficient way to test for local dev...?

Rob--W: I recommended against `about:debugging` before because you wanted to test a browser restart, but the extension would be uninstalled due to it being temporary.

If you don't restart the browser, then it is OK to use `about:debugging` or `web-ext` + R (reload).

When you disabled signature checking (see my previous comment in this thread), then you don't need to use `about:debugging` or `web-ext` to reload the extension. You can simply open the `.xpi` file in Firefox to install it. This experience is much closer to what the user will do in practice.

rebmullin: > To test the behavior of an add-on after a browser restart, you have to actually install an add-on...

how would i go about installing the addon w/out using about:debugging /addons page? I tried dragging the unsigned dev xpi to the browser but not sure if that's working since I still see "Get Firefox Color" button.

(note: I have set `xpi.signatures.required` to false)

Rob--W: > how would i go about installing the addon w/out using about:debugging /addons page? I tried dragging the unsigned dev xpi to the browser

That should work.

> but not sure if that's working since I still see "Get Firefox Color" button.

Did you get an install prompt? If so, it did work, but it could be that you didn't install the right xpi file. Where did you get your xpi file?

For `localhost`, you should run `npm run package` and use the generated `addon.xpi`. You can open the directory as a `file:///`-URL in your browser and click on `addon.xpi` to install it. If this is not obvious (which I'm afraid is the case), consider adding it to the README in your existing attempt to improve README (this is optional, you don't have to do it).

rebmullin: huh?! 
yeh I didn't know about `npm run package` (I just haven't needed it yet). I noticed after running this script an `addon.xpi` gets created which I saw gets added to the root of the project. I'll definitely update the README so it's more clear for others that are working on this project. thanks!

rebmullin:
hey @Rob--W, can you please take another look at this? 

notes: when you initially enable the addon, it loads the addon’s default theme. After hitting “Revert All,” FF's default theme is used and the addon’s default theme (the purplish/gray color) will not be activated again. The user must start editing (via "Random," etc actions) to "activate" the addon again - which will then set the theme. If the user has a url from share for example, it will also activate the addon so the theme is displayed.  If the last action was “Revert All” and the user disables the addon and then enables later on, the same activation rules apply as noted above (the user must start interacting with the UI to "activate").
If the user updates the theme - but has not hit “Revert All, ”  and then disables the addon and then re-enables the addon,  the last theme set will then be shown.

src/lib/themes.js
```diff
@@ -160,6 +160,7 @@ export const normalizeThemeColors = (colors = {}) => {
 
 // Utility to ensure normal properties and values in app theme state
 export const normalizeTheme = (data = {}) => {
+  if (!data) return undefined;
```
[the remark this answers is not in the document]
rebmullin: hmm I did try to avoid setting this and thought I saw an error but I am not seeing it now :/ 🤷 

I'll revert for now.

package.json
```diff
@@ -157,7 +157,8 @@
     "permissions": [
       "theme",
       "storage",
-      "tabs"
+      "tabs",
+      "management"
```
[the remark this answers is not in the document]
rebmullin: ahh ok, I misunderstood. I thought you suggested that I should do this here. So that will definitely simplify things i think. I had this before actually but then when i re-read the tickets/comments I changed it 🤷 

So anyway, just to clarify hitting "Revert All" will default back to FF's default theme (regardless of what FF theme is set)

rebmullin:
hey @Rob--W, how's it look now? :)

src/extension/background.js
```diff
@@ -136,6 +146,14 @@ const storeImages = ({ images }) => browser.storage.local.set({ images });
 const BLANK_IMAGE =
   "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=";
 
+
+  browser.storage.onChanged.addListener(ui => {
+    if (!ui || ui.hadUIInteraction && !ui.hadUIInteraction.newValue) {
```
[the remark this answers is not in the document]
rebmullin: I tried that but it wasn't working as expected. I'll revisit though to see if I can remember why :)

src/lib/themes.js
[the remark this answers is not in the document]
rebmullin: sorry i thought I reverted this : /

src/extension/background.js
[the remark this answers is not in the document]
rebmullin: seems to be okay with the other changes so I'll move this. thanks!

src/web/lib/middleware.js
```diff
@@ -11,20 +11,35 @@ export default function({
...
   const updateHistoryMiddleware = ({ getState }) => next => action => {
     const returnValue = next(action);
     const meta = action.meta || {};
+
+    const { search } = window.location;
+
+    // If the user does anything with a theme or the url has any params (for firstRun or share for eg), 
+    // the theme will be activated.
+    if (meta.userEdit || Boolean(search)) {
+      postMessage("activateExtension");
```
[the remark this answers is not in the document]
rebmullin: hmm I am not sure of another way to have the url shared and have it start right away w/o this...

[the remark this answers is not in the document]
rebmullin: for now i removed this for now but gonna think about this 1 some more..

src/web/index.js
```diff
@@ -154,13 +154,17 @@ window.addEventListener("message", ({ source, data: message }) => {
         postMessage("addImages", {
```
Rob--W, line 154: Move this inside the `hasEdited` condition as well.

src/web/lib/middleware.js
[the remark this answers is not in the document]
Rob--W: There should already be a prompt to allow the user to load the shared theme. The expected flow is documented here:

https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L247-L270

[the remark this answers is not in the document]
rebmullin: ok cool, it seems to work if I adjust the meta when ?theme is present..
so I made a little update there too.

src/web/index.js
```diff
@@ -290,7 +295,7 @@ if (!params.theme) {
           // Skip updating history for this theme, because it came from the URL
           skipHistory: true,
           // Skip updating the add-on for this theme, because it needs approval
-          skipAddon: true
+          skipAddon: false
```
Rob--W, line 298: Why was this change made?

This line of code contradicts the code comment above. With this proposed change, opening a shared theme will unconditionally override the user's current theme without prompting. That's undesired.

When a shared theme is opened, there should be some UI from where the user can confirm that they want to apply the new theme. This is implemented in the "pending theme" feature in Firefox Color. If the prompt does not appear as expected, I suggest to look into its implementation to try and figure out why.

PR #851, opened 2019-10-14
Fixes #846

### LINKED ISSUE #846: Dark icon
In the Firefox addons dark page, the FirefoxColor icon is hard to see, will be great make a medium gray or something like that.

Dark Icon2

ShivamRawat0l:
Set the icon for the dark theme .

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,6 +159,27 @@
       "storage",
       "tabs"
     ],
+    "theme_icons": [{
+      "light": "images/logo.svg",
+      "dark": "images/logo-white.svg",
+      "size": 16
+    }, {
+      "light": "images/logo.svg",
+      "dark": "images/logo-white.svg",
+      "size": 32
+    }, {
+      "light": "images/logo.svg",
+      "dark": "images/logo-white.svg",
+      "size": 48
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -159,6 +159,27 @@
       "storage",
       "tabs"
     ],
+    "theme_icons": [{
```
Rob--W: `theme_icons` is not a top-level property, but part of `browser_action`. Since it's a vector image (SVG), it suffices to specify just one size: 32.

package.json
```diff
@@ -182,17 +203,11 @@
         "run_at": "document_start"
       }
     ],
-    "icons": {
```
Rob--W: Keep this. `theme_icons` only applies to the browserAction button.

Rob--W:
Note in case it's not clear: This patch does not resolve #846 , because it is currently not possible to have different icons in the add-on manager depending on the theme.

ShivamRawat0l:
@Rob--W   Yes I just found that. It only changes the color of the icon in the toolbar. Thanks for helping.

PR #857, opened 2019-10-20
I852 adding cookie for modal tracking

### LINKED ISSUE #852: The user is not informed that deleting an image used in a saved theme will also be deleted from the saved theme
**[Notes]:**
- As this issue is not reproducible on the [production website](https://color.firefox.com) and existing users are already used that deleting an image used in saved themes from the “Custom backgrounds” section will not delete the image from the saved themes, this new behavior might seem like data loss. 
- By informing the users that there is a new behavior when deleting an image used in saved themes and asking for confirmation in order to proceed with deleting the image might avoid any confusion. 

**[Affected versions]:**
- Firefox Nightly 71.0a1, Build ID: 20191013213650
- Firefox Beta 70.0b14, Build ID:20191013213650
- Firefox Release 69.0.3, Build ID: 20191009172106

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox Color add-on zip modified in order to link to the [stage](https://color.stage.mozaws.net/) website.
- Have the previous Firefox Color add-on installed through about:debugging.
- Have a saved theme containing an image.

**[Steps to reproduce]:**
1. Open the profile from the prerequisites.
3. Select the saved theme and go to the “Custom backgrounds” section.
4. Click the “Delete” button from the image.
5. Observe the behavior.

**[Expected result]:**
- A modal is displayed informing the user that deleting the image will also delete it from the saved theme and asking the user for permission in order to proceed.

**[Actual result]:**
- The image is deleted.

rebmullin:
fixes #852 

hey @Rob--W, can you please do a review on this 1 if you have some time?

hey @caitmuenster, if you want to take a look now too or wait till Rob signs off.. 

note: here is a "sketch" that I got approval on:

thanks:)


src/web/lib/components/ClearImageModal/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+    this.props.setDisplayRemoveImageModal({ display: false });
+  }
+
+  render() {
+    return (
+      <Modal toggleModal={this.toggalModal} displayModal>
+        <p>Deleting this image will remove it from any saved themes you have. Do you want to proceed?</p>
+        <div className="modal__buttons">
+          <button onClick={this.toggalModal.bind(null, { display: true })}>Okay</button>
```
Rob--W: Consider splitting `toggalModal`, e.g. one method called `confirmRemove` and the other called `cancelRemove`.

caitmuenster:
Super small nit -- there should be a ? at the end of the "Do you want to proceed?"

rebmullin:
hey @caitmuenster, here is the updated screenshot (with the ? at the end):

rebmullin:
hey @Rob--W, I appreciate the feedback. I've made some changes per your suggestions. Please let me know how it looks now. thanks!

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -298,6 +298,8 @@ class ThemeCustomBackgroundSelector extends React.Component {
 
               <ImportButton label={errors ? "Retry" : "Replace image"} />
 
+              {this.props.displayRemoveImageModal && <div className="modal-wrapper--clear-image"><ClearImageModal {...this.props} /></div>}
```
Rob--W: Rather than blindly forwarding `this.props` and duplicating the names "clearImageModal", `clearCustomBackground`, `setDisplayRemoveImageModal` in two different files, would it be possible to keep the logic in one place?

For example, the modal file:

```
<button onClick={this.props.confirm}>...
<button onClick={this.props.cancel}>...
```

and here:

```
<ClearImageModal
  confirm={() => this.props.clearCustomBackground() }
  cancel={() => this.props.setDisplayRemoveImageModal({ display: true }) }
>
```

rebmullin: hey @Rob--W , I definitely like the idea of keeping this in one 1 file I just thought it was better to set the localStorage key `clearImageModal` in the modal itself. If I change it, I would have something like this in the ThemeCustomBackgroundPicker file:

```javascript
  <ClearImageModal
     confirm={this.confirm}
     cancel={this.onCloseModal}
  />

...

  confirm = () => {
    this.props.clearCustomBackground();
    this.onCloseModal();
  }

  onCloseModal = () => {
    // the following needs to happen on "okay" or "cancel"
    this.props.setDisplayRemoveImageModal({ display: false }); 
    localStorage.setItem("clearImageModal", true);
  }

  handleClearBackground = () => {
    if (localStorage.getItem("clearImageModal")) {
      this.props.clearCustomBackground();
    } else {
      this.props.setDisplayRemoveImageModal({ display: true });
    }
  };
```

if you agree that setting the `clearImageModal` localStorage value here is fine (instead of in the `ClearImageModal` component), then I'll go ahead and move this as shown.

rebmullin: Otherwise, I'll keep it as I currently have it. but wanted to know your thoughts before I changed/merged anything :)
thanks!

Rob--W: The snippet that you described looks good to me.

caitmuenster:
> hey @caitmuenster, here is the updated screenshot (with the ? at the end):
> 
> 
> 
> how's it look now? :)

This looks great! Thank you!

caitmuenster:
@Rob--W sorry for the ping, I know you're out this week! Are you able to give this patch an r+ or should we ask Luca?

PR #854, opened 2019-10-19
I853 disable undo after bg delete

rebmullin:

Issue #853, opened 2019-10-15
A blank custom background item is wrongly displayed after pressing the “Undo” button if an image was previously deleted


Softvision-AndreeaCupsa:
**[Affected versions]:**
- Firefox Nightly 71.0a1, Build ID: 20191013213650
- Firefox Beta 70.0b14, Build ID:20191013213650
- Firefox Release 69.0.3, Build ID: 20191009172106

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox Color add-on zip modified in order to link to the [stage](https://color.stage.mozaws.net/) website.
- Have the previous Firefox Color add-on installed through about:debugging.

**[Steps to reproduce]:**
1. Open the profile from the prerequisites.
2. Navigate to the https://color.stage.mozaws.net/ website.
3. Click the “Custom backgrounds” section and add a valid image.
4. Click the “Delete” button from the image.
5. Click the “Undo” button from the top toolbar.
6. Observe the “Custom backgrounds” section.

**[Expected result]:**
- The section of the previously deleted image is not displayed.

**[Actual result]:**
- The image background section is displayed but it does not contain an image.

**[Notes]:**
- This issue is not reproducible on the [production website](https://color.firefox.com). Here the image background section is correctly displayed with all its elements.
- Attached a screen recording with the issue: 
![Delete image](https://user-images.githubusercontent.com/37875464/66832165-00c11c80-ef62-11e9-93ab-cba6577618d7.gif)

caitmuenster:
Thanks, @Softvision-AndreeaCupsa! @rebmullin, is it ok if I ask you to take a look at this?

Softvision-AndreeaCupsa:
I have verified this issue and is no longer reproducible using the production xpi (version 2.1.5) linked to the stage server on Firefox Release 72.0.1 (Build ID:20200107212822), Firefox Beta 73.0b2 (Build ID: 	20200107212705) and Firefox Nightly (Build ID: on Windows 10 x64, Mac 10.14.6 and Ubuntu 18.

PR #868, opened 2019-11-18
fix delete image from local storage

rebmullin:


caitmuenster:
PR fixes a regression where deleting images or themes from the UI did not remove them from local.storage. 

Hey @Rob--W, could you review this?

src/web/lib/components/SavedThemeSelector/index.js
```diff
@@ -8,21 +8,57 @@ export const SavedThemeSelector = ({
...
+
+      const notUsedImages = currentImageNames.filter(image => !allThemeImages.includes(image));
+
+      notUsedImages.forEach((image) => {
+        clearCustomBackground({ index: 0 });
+        imageStorage.delete(image);
+      });
+    }
+
+    themeStorage.delete(key);
```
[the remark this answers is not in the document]
rebmullin: I think I did look at this but I'll recheck..

src/web/lib/components/SavedThemeSelector/index.js
```diff
@@ -8,21 +8,30 @@ export const SavedThemeSelector = ({
...
+  setCurrentThemeId
 }) => {
   const { themeStorage } = storage;
   const sortedSavedThemes = Object.entries(savedThemes).sort(
     ([, aData], [, bData]) => bData.modified - aData.modified
   );
 
+  const deleteTheme = (key) => {
+    // This is so we can track which theme to delete in some other places.
+    setCurrentThemeId({ key });
```
Rob--W: Why is this needed? I expect that `selectors.savedThemes(state)` already returns themes without the deleted theme, simply because it is not in the storage any more. Is the theme somehow returning without this hack?

When `themeStorage.delete` is called, the "storage" event is triggered, which updates the saved themes already: https://github.com/mozilla/FirefoxColor/blob/17092506ea355f49894eafe406a675cc114af9f3/src/web/lib/storage.js#L147-L150

One thing that seems to be missing is that `currentSavedTheme` is still at the old value. I suppose that it may be necessary to call `this.afterPut(undefined);` to properly dispose of the cleared theme in `delete` at
https://github.com/mozilla/FirefoxColor/blob/17092506ea355f49894eafe406a675cc114af9f3/src/web/lib/storage.js#L55

rebmullin: hey, thanks again for the feedback :)

This is a bit trickier than I thought but I'll try to take a look over the next few days.. But if I recall, the theme was still in there which is why I approached it this way. I'll take another look at using afterPut. maybe that's the key..? I'll report back :)

rebmullin:
hey @Rob--W, I reverted some of the work (it seems to be working as expected now). can you please take another look?

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -327,12 +326,19 @@ class ThemeCustomBackgroundSelector extends React.Component {
...
     } else {
       this.props.setDisplayRemoveImageModal({ display: true });
     }
   };
 
+  removeImage = () => {
+    // TODO: update / remove /resave theme/s without this image.
+    const { image } = this.props;
+    this.props.clearCustomBackground();
+    this.props.storage.imageStorage.delete(image.name);
```
Rob--W: The image is only present because it is in one of the entries in the `themePast` array at https://github.com/mozilla/FirefoxColor/blob/17092506ea355f49894eafe406a675cc114af9f3/src/web/lib/middleware.js#L86

This was a consciously added change in #418, as part of a fix to fix #414.
When we are not interested in the image, then we might as well remove the code from `purgeImages` in middleware.js, starting from `// Scan through undo/redo stack for images still in use.` until the end, i.e. https://github.com/mozilla/FirefoxColor/blob/17092506ea355f49894eafe406a675cc114af9f3/src/web/lib/middleware.js#L85-L96

Could you check whether that would also fix your bug and still not regress on #414 ?

rebmullin: This is what we want: we want to remove the background image from local storage when the user clicks the X here (the user is warned about this the first time they try to do this. this was the purpose of #852). so yes it looks like I can remove this logic: `// Scan through undo/redo stack for images still in use...`  
Regarding “undo/redo”, I tried to change this with #854 but unfortunately I do still see a bug here. So I can try to remove this action from the undo/redo action list or maybe set up another action to flag this but I’ll have to look into this more..

caitmuenster:
Hey @Rob--W, could you review the latest commit by Dec 6 at 12pm PST? Ideally we'd like to push tomorrow and we're gated on this.

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
rebmullin: hey @Rob--W. I tried to skip this action per the undo/redo instructions but it wasnt working as expected so I adjusted the userEdit value for this and it *seems* to be ok. But this is what we want - to remove the image from local storage.

Rob--W:
What is the main objective of the PR? I thought that it was to remove the image from the storage, but the latest version also makes other changes that I don't understand (which could very well be my limited familiarity with the libraries here).

Are you trying to make "remove background" not part of the undo/redo stack? If so, have you tried moving it out of the `undoable` function?

caitmuenster:
Thanks @Rob--W. This PR should fix some previous regressions that resulted in images and themes not being removed from local storage. Here's a quick summary of expected behavior: 

### Custom Background Image Removal Flow

* User clicks the X on a custom background / gets prompted asking if it's ok to remove image / confirms 
* Custom background image is removed from theme
* Color checks every saved theme to see if background image is used
* If no other saved theme references this custom background image, the custom background image is deleted from local storage 

### Theme Deletion Flow
* User deletes a saved theme
* Color checks each custom background image used by that theme and looks to see if other saved themes use those custom background images
* If no other saved theme references those custom background image,  the custom background images are deleted 

### Undo 
* Users should not be able to click "Undo" after deleting a custom background image. (Trying to "Undo" the deletion of a custom background image issue #853.)

rebmullin:
> Are you trying to make "remove background" not part of the undo/redo stack? If so, have you tried moving it out of the undoable function?

hey @Rob--W, I tried this the following ` if (themesEqual(previousHistory.present, currentState) && action.type === "CLEAR_CUSTOM_BACKGROUND") {...` (per https://www.npmjs.com/package/redux-undo#filtering-actions) (I noted ~ this above) but it wasn't working as expected. I guess we can remove this userEdit piece and create a new issue/PR if this is a concern but it seems to resolve the image issue you noted above (in #414).

rebmullin:
hey @Rob--W, can you please take another look at this one? I've reverted the updates for undo/redo here and I am only addressing the local storage issue (to delete the image from local storage once the user clicks "X"). I will work on the redo/undo (using global var approach for the image) in a new PR.

// cc @caitmuenster

PR #872, opened 2019-12-10
adding storedimages map to help w/undo/redo

rebmullin:


src/web/lib/middleware.js
```diff
@@ -1,4 +1,5 @@
 import { actions, selectors, themeChangeActions } from "../../lib/store";
+export let storedImages = new Map();
```
Rob--W: To emphasize that this is not persistent, but a memory-only storage area, rename it to `temporaryImageStore`.

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -156,23 +162,43 @@ const BackgroundList = SortableContainer(props => {
...
+    let image = storedImages.get(name);
+
+    if (image) {
+      // eslint-disable-next-line no-unused-vars
+      const { importing, ...rest } = image;
+
+      const newImage = {
+        ...rest
+      };
+      // add the image back into local storage.
```
Rob--W: When would this happen?

I'm fine with automatically adding the image back upon redo/undo, but then when the user use redo/undo again, then the image should be deleted again.

rebmullin: this adds the image if its looking for it on undo/redo and the image is missing.

ok thanks for the feedback. I'll look into this..

rebmullin:
hey @Rob--W. this is not quite ready for another review. I'll let you know when it is:)

rebmullin:
hey @Rob--W, please take another look when you can. thanks!

caitmuenster:
Thanks @rebmullin! Biggest issue I saw on my local test was when I clicked on an image to delete it, a different image was deleted instead. Video is here: https://drive.google.com/file/d/1UKgigeIzaIz0-co4os_BtyFLkzsEaAI2/view?usp=sharing). 

Otherwise, undo/redo worked ok.

rebmullin:
> Biggest issue I saw on my local test was when I clicked on an image to delete it, a different image was deleted instead.

oh?! hmm..let me take another look..

rebmullin:
hey @caitmuenster, thanks for the feedback. I've made some changes. can you please take another look? thanks so much:)

caitmuenster:
Awesome, thanks @rebmullin! Image deletion, undo, and redo all work as expected ([video](https://drive.google.com/drive/folders/1R6peYr-9ExAUXHL0A_h92fdY98Wdo7EY)). 🙌🏼

(As a quick reference note, undo'ing the deletion of a background image seems to only work from the "Custom Backgrounds" screen (and not from the "Saved Themes" screen, which I'm fine with.)  

@Rob--W, could you review the code?

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -23,6 +24,37 @@ export class ThemeCustomBackgroundPicker extends React.Component {
     super(props);
   }
 
+  componentDidUpdate(prevProps) {
+    const { themeCustomBackgrounds } = this.props;
+
+    // sync images in local storage with custom backgroundsin preview on undo/redo
+    if (prevProps.themeCustomBackgrounds.length !== themeCustomBackgrounds.length) {
```
Rob--W: This condition (and the conditions below that also rely on lengths) only works if an item is added, removed or re-ordered.
If an item is replaced, or if multiple items are added/removed and re-ordered, then the logic here may not work as expected (i.e. images being added/removed while they should not be).

Are these length-conditions correct?

rebmullin:
hey @Rob--W, thanks so much for checking. I'll have to take another look tho. honestly, this one has been a bit tough but thought it was "okay" b/c it was working 😬 

Anyway, I won't have a chance again to recheck this till over the weekend/next week. I'll ping you when it's ready or if any questions (I do realize you both might be out b/c of the holiday).

//cc @caitmuenster

caitmuenster:
It's totally ok with me! We weren't going to release this week due to folks being out for the holidays. :) @rebmullin if you could take a look at it this weekend, we can get it queued up for stage/QA sometime the week of the 30th (although we will be trying to get back into the post-New Year routine).

rebmullin:
Since there was a little bit of back and forth about this, I wanted to make some notes about the expectations on adding/removing from local storage.

If this should be posted somewhere else or removed please let me know:) Also, please let me know if there are any questions or if I have left anything out:

--------

- click “X” on the "Custom backgrounds" tab

image is deleted from local storage. Currently the theme will still mention the background image (if it was saved as part of it) but it will not be exported with this image (or show in image preview with this image). Also - as the modal warns - this will also be deleted from any saved themes that references it.


- add custom background image on the "Custom backgrounds" tab

image is added to local storage


- click "Undo"/"Redo" on “X” or add custom background steps

image/s will be added /removed from local storage respectively (if seen in custom background list, it will be added. if not in the preview list, it will be removed *if* not saved in any themes).


- click “X” theme on "Saved themes" tab

This will remove the background image/s from local storage *if* the background image/s in this theme are not used in any another theme. Otherwise, the image/s will be removed from local storage.


additional notes on local storage cleanup (i.e. an image from local storage will be removed in the following scenarios):
-if you add an image without saving this to a theme, it will be removed upon redo/undo  (i.e. *if* it's not part of the adding/removing step:)
-if you add an image without saving this to a theme, it will be removed upon selecting another theme (via “Random” or selecting a preset theme)
-if you delete a theme, it will scan through unused background images in local storage
-reload (*if* it's not in current preview, even if it's not saved)

----

caitmuenster:
Thanks @rebmullin! This looks good on local. 

Color - delete undo logic

@Rob--W, could you take a look when you're back from the holidays?

src/web/lib/components/AppHeader/index.js
```diff
@@ -45,29 +54,69 @@ export const AppHeader = props => {
...
-      </button>
-      {children}
-    </React.Fragment>
-  );
+
+  ) => {
+    const handleButtonClick = (event) => {
+      const { currentTarget: { title } } = event;
+      onClickButton();
+      setUpdate(title === "Undo" || title === "Redo" || title === "Random");
```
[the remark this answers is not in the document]
rebmullin: I believe it's to exclude buttons..

rebmullin:
Hi @caitmuenster + @Rob--W, thanks so much for the feedback :)

I made some changes per the comments. Unfortunately, I did not have a ton of time to re-test this thoroughly (I can do this possibly Sunday but most likely Monday/Tuesday evening next week).

caitmuenster:
I think this is still looking ok -- here's a [video](https://drive.google.com/open?id=128uxkZEyTSaBZQ0LkkzpzDRNfOP2fKPb) of his afternoon's test. 

@Rob--W r?

src/web/lib/components/AppHeader/index.js
```diff
@@ -45,29 +54,69 @@ export const AppHeader = props => {
...
-          <img src={icon} width="20" height="auto" aria-hidden="true" />
-        </div>
-        <span>{text}</span>
-      </button>
-      {children}
-    </React.Fragment>
-  );
+
+  ) => {
+    return (
```
Rob--W: nit: The diff can be smaller by undoing the addition of `return (` and whitespace.

Issue #361, opened 2018-06-20
Allow users to preview shared themes.


sevaan:
When viewing a shared theme, hovering over the preview image should allow the user to preview it in their browser chrome before they apply it.

<img width="602" alt="screen shot 2018-06-19 at 3 07 51 pm" src="https://user-images.githubusercontent.com/473531/41667811-e1bc049c-747b-11e8-8dc9-3172c86d482e.png">

lmorchard:
Some implementation notes off the top of my head:

* On the extension side, we'll need to accept a message to set & clear a temporary theme such that it:
  1. does *not* get saved to local storage
  1. gets replaced with the last chosen theme on clear
  1. gets replaced with the next chosen theme, in case we somehow miss clearing it

* On web side in the preview component, send a message to set temporary theme on rollover. On rolloff, clear temporary theme.

Because this is so ephemeral, we can probably skip adding temporary theme support to the Redux store on the web client. We don't need to track it in undo/redo or update web app state. 

If we end up wanting to display some feedback in the web app on rollover, then we probably should add it to Redux store so that React components can use it.

Shivansh2407:
Can I please work on this issue and also please explain me from where should I start?

rebmullin:
hey @ValentinaPC, is it possible to QA this? It should be on stage to be reviewed?

PR #884, opened 2020-02-06
I361 preview shared theme on rollover

r-ba:
The intent of this PR is to add a rollover effect to the preview image in the `SharedThemeDialog` component as proposed in #361.

### DIFF
--- a/src/extension/background.js
+++ b/src/extension/background.js
@@ -62,11 +62,11 @@ const messageHandlers = {
     storeTheme({ theme });
     applyTheme({ theme });
   },
-  previewTheme: message => {
-    if (message.theme) {
-      const theme = normalizeTheme(message.theme);
-      log("previewTheme", theme);
-      applyTheme({ theme });
+  previewTheme: ({ theme }) => {
+    log("previewTheme", theme);
+    if (theme) {
+      const previewTheme = normalizeTheme(theme);
+      applyTheme({ theme: previewTheme });
       isThemePreview = true;
     } else {
       fetchTheme().then(applyTheme);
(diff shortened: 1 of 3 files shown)


src/extension/background.js
```diff
@@ -51,11 +56,20 @@ const messageHandlers = {
...
     applyTheme({ theme });
   },
+  previewTheme: message => {
+    if (message.theme) {
+      const theme = normalizeTheme(message.theme);
+      log("previewTheme", theme);
+      applyTheme({ theme });
+      isThemePreview = true;
+    } else {
+      fetchTheme().then(applyTheme);
```
r-ba: Should there be a `log` here? E.g. `log("revertTheme");`

r-ba:
Hi @Rob--W, thanks for the feedback.

Out of curiosity, is reverting the theme via a `port.onDisconnect` listener meant to act as a fail-safe? In the previous implementation if a user closes the tab with their cursor or a keyboard shortcut the theme seemed to be reverted as expected. Just wondering if I'm missing something more here.

r-ba:
> Did it? From looking at the code diff, it seemed that the theme wouldn't be reset if the tab was closed/navigated elsewhere without moving the mouse.

Yes it did, surprisingly enough. In all honestly I only have a vague guess as to why that is.

src/extension/background.js
r-ba: Done. A few notes to consider: 
- The onMouseLeave event sends an empty object via `postMessage`, effectively telling the extension to revert to the stored theme. When this occurs the `theme` argument passed to `log` will be `undefined`, such that the console log will be:
`[FirefoxColor background] previewTheme undefined`. 
- In the `setTheme` message listener, perhaps we may wish to also destructure `theme` from `messages` for the sake of consistency.

PR #890, opened 2020-03-01
I881 updating pg when deleting last saved theme on pg

### LINKED ISSUE #881: Last page of "Saved themes" tab is not removed after you delete all listed themes
**Affected versions:**
Firefox Nightly 74.0a1 - Build ID 20200109213942

**Affected Platforms:**
All Windows
All Mac
All Linux

**Steps to reproduce:**
1. Navigate to color.firefox.com 
3. Have in "Saved themes" tab, more than 3 pages. (an easy way to do this is to randomly generate and save themes)
4. Navigate to last page and delete all themes from that page.

**Expected results:**
User is redirected to previous page.
Navigation section shows active page.

**Actual result:**
A blank page is displayed. Navigation section does not show an active page.

**Notes:**
This issue is also reproducible after "Firefox Color" add-on installation.
Attaching a screenshot with the issue.
delete themes

rebmullin:
fixes #881

### DIFF
--- a/src/web/lib/components/PaginatedThemeSelector/index.js
+++ b/src/web/lib/components/PaginatedThemeSelector/index.js
@@ -8,14 +8,6 @@ import "./index.scss";
 
 import iconClose from "./close.svg";
 
-const usePrevious = value => {
-  const ref = React.useRef(null);
-  React.useEffect(() => {
-    ref.current = value;
-  });
-  return ref.current;
-};
-
 export const PaginatedThemeSelector = ({
   themes,
   className,
@@ -28,12 +20,11 @@ export const PaginatedThemeSelector = ({
   images = []
 }) => {
   const itemCount = themes.length;
-  const previousCount = usePrevious(themes.length);
 
   React.useEffect(() => {
-  if (itemCount !== previousCount && previousCount !== null && itemCount % perPage === 0 && currentPage !== 0) {
-    setCurrentPage(currentPage - 1);
-  }
+    if (itemCount && itemCount % perPage === 0 && currentPage) {
+      setCurrentPage(currentPage - 1);
+    }
  }, [itemCount]);
 
   const pageCount = Math.ceil(itemCount / perPage);
--- a/src/web/lib/components/PaginatedThemeSelector/index.js
+++ b/src/web/lib/components/PaginatedThemeSelector/index.js
@@ -8,6 +8,14 @@ import "./index.scss";
 
 import iconClose from "./close.svg";
 
+const usePrevious = value => {
+  const ref = React.useRef(null);
+  React.useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
+
 export const PaginatedThemeSelector = ({
   themes,
   className,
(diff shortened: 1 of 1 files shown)


src/web/lib/components/PaginatedThemeSelector/index.js
```diff
@@ -20,6 +28,14 @@ export const PaginatedThemeSelector = ({
   images = []
 }) => {
   const itemCount = themes.length;
+  const previousCount = usePrevious(themes.length);
+
+  React.useEffect(() => {
+  if (itemCount !== previousCount && previousCount !== null && itemCount % perPage === 0 && currentPage !== 0) {
```
Rob--W: Indentation looks off.

Do you really need the `usePrevious` hook? Wouldn't the issue be fixed if you use
```
if (itemCount && currentPage >= itemCount) {
  setCurrentPage(itemCount - 1);
}
```

(I don't know if it is possible, but if there was a bulk-delete functionality, then the current proposal would fail again, whereas the proposed snippet would not break)

rebmullin:
hey @Rob--W, I took out usePrevious hook. let me know what you think now. thanks :)

src/web/lib/components/PaginatedThemeSelector/index.js
```diff
@@ -20,6 +20,13 @@ export const PaginatedThemeSelector = ({
   images = []
 }) => {
   const itemCount = themes.length;
+
+  React.useEffect(() => {
+    if (itemCount && itemCount % perPage === 0 && currentPage) {
```
Rob--W: Is this logic correct? If there is only one item per page (`perPage = 1`), then  `itemCount % perPage === 0` would always be true. As a result, whenever this logic is triggered, the current page would decrease by one, which doesn't seem right to me.

Did the condition here not work? https://github.com/mozilla/FirefoxColor/pull/890#discussion_r387768166

rebmullin: it not did not work.

```javascript
if (itemCount && currentPage >= itemCount) {
  setCurrentPage(itemCount - 1);
}
```

if currentPage is 2 (page 3) but itemCount is 19, then it will never switch pages when itemCount gets changed to 18 (in this case we want to switch to currentPage to 1 (page 2) 

it seems that perPage can't be changed via the ui tho, right? so i wasn't accounting for perPage to ever be 1 (or different than 9).

Rob--W: Ow facepalm. Need to multiply or divide by perPage to be able to meaningfully compare currentPage and itemCount.

rebmullin: it seems that what I have works or I can slightly adjust what you did above to:

```javascript
 if (itemCount && currentPage >= (itemCount / perPage) ) {
    setCurrentPage(currentPage - 1);
 }
```

Rob--W: That looks almost right, but only when either one theme is being deleted (which is often the case in practice).

If there was a delete-all feature, and the user is on the last page prior to deleting all themes, then the current page would be set to a non-existing page. To fix that, I believe that the logic should be:

```
let pageCount = Math.ceil(itemCount / perPage);
if (itemCount && currentPage >= pageCount) {
  setCurrentPage(pageCount - 1);
}
```

PR #891, opened 2020-03-01
I888 fixing tooltip icon error

### LINKED ISSUE #888: The tooltips of the Warning icon next to the uploaded image are incorrect after trying to upload incompatible files
[Affected Platforms]:
- Windows 10 x64
- Ubuntu 16.04 x64
- macOS 10.14

[Affected Versions]:
- Firefox Nightly 75.0a1 (2020-02-19) Build ID 20200219095352
- Firefox Beta 74.0b5 , Build ID 20200218224219 
- Firefox Release 73.0.1 Build ID 20200207195153

[Prerequisites]:
- Have the latest version of the Firefox Color add-on modified in order to link to the stage website.
- Have the previous Firefox Color add-on installed through "about:debugging".

[Steps to reproduce]:
3. Navigate to "Custom backgrounds" tab and click the "...Or add your own" button.
4. Upload a picture up to 1MB (jpg, png or bmp).
5. Click the "Replace image" button.
6. Upload a file bigger than 1MB.
7. Hover the Warning icon from the left side of the image's thumbnail.
8. Click the "Replace image" button.
9. Upload a file that is not an image (eg. xpi, zip, pdf, js etc).
10. Hover the Warning icon from the left side of the image's thumbnail.

[Expected result]:
- Step 7 - The Warning tooltip contains the same text as the error message "The image is too large (1MG maximum size)" OR nothing happens.
- Step 10 - The Warning tooltip contains the same text as the error message "The file is not an accepted image type" OR nothing happens.

[Actual result]:
- Step 7 - {"too large":true, "wrongType":false} tooltip is displayed.
- Step 10 - {"too large":false, "wrongType":true} tooltip is displayed.

[Notes]:
- I have attached a screen recording of the issue:

color

rebmullin:
fixes #888

### DIFF
--- a/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
+++ b/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
@@ -20,10 +20,15 @@ import iconHAlignLeft from "./icon_align_left.svg";
 import iconVAlignCenter from "./icon_align_center.svg";
 
 
-const mapErrors = (errors) => {
-  if (errors.tooLarge) {
+const mapErrors = ({ tooLarge, wrongType}) => {
+  if (tooLarge && wrongType) {
+    return `
+    The image is too large. (1MB maximum size) 
+    The file is not an accepted image type.
+    `;
+  } else if (tooLarge) {
     return "The image is too large. (1MB maximum size)";
-  } else if (errors.wrongType) {
+  } else if (wrongType) {
     return "The file is not an accepted image type.";
   }
   return "";
@@ -138,9 +143,9 @@ export class ThemeCustomBackgroundPicker extends React.Component {
                 {errors && (
                   <React.Fragment>
                     <Modal>
-                      <ul className="errors">
-                        <li>{mapErrors(errors)}</li>
-                      </ul>
+                      <div className="errors">
+                        <span>{mapErrors(errors)}</span>
+                      </div>
                     </Modal>
                   </React.Fragment>
                 )}
--- a/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
+++ b/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
@@ -19,6 +19,16 @@ import "./index.scss";
 import iconHAlignLeft from "./icon_align_left.svg";
 import iconVAlignCenter from "./icon_align_center.svg";
 
+
+const mapErrors = (errors) => {
+  if (errors.tooLarge) {
+    return "The image is too large. (1MB maximum size)";
+  } else if (errors.wrongType) {
+    return "The file is not an accepted image type.";
+  }
+  return "";
+};
+
 export class ThemeCustomBackgroundPicker extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -129,12 +139,7 @@ export class ThemeCustomBackgroundPicker extends React.Component {
                   <React.Fragment>
                     <Modal>
                       <ul className="errors">
-                        {errors.tooLarge && (
-                          <li>The image is too large. (1MB maximum size)</li>
-                        )}
-                        {errors.wrongType && (
-                          <li>The file is not an accepted image type.</li>
-                        )}
+                        <li>{mapErrors(errors)}</li>
```
Rob--W: Previously, more than one error message could be shown.

With this change, only one message is shown.

rebmullin: hmm i see..I'll revisit. thanks!

rebmullin:
hey @Rob--W, how is it now?

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -19,6 +19,21 @@ import "./index.scss";
 import iconHAlignLeft from "./icon_align_left.svg";
 import iconVAlignCenter from "./icon_align_center.svg";
 
+
+const mapErrors = ({ tooLarge, wrongType}) => {
+  if (tooLarge && wrongType) {
+    return `
```
Rob--W: This function is used both for generating the HTML and title attribute. HTML is insensitive to line breaks (treated as whitespace), but the title attribute is not. I think that with this change, you may end up with a tooltip that has a leading and trailing blank line.

To account for this, how about returning an array here, and let the callers format the message as they see fit? For example:

HTML:

```
mapErrors(errors).map(msg => `<li>${msg}</li>`);
```

Title:
```
mapErrors(errors).join("\n")
```

rebmullin: yep :) i agree that's a lot better!

src/web/lib/components/ThemeCustomBackgroundPicker/index.js
```diff
@@ -129,12 +141,7 @@ export class ThemeCustomBackgroundPicker extends React.Component {
                   <React.Fragment>
                     <Modal>
                       <ul className="errors">
-                        {errors.tooLarge && (
-                          <li>The image is too large. (1MB maximum size)</li>
-                        )}
-                        {errors.wrongType && (
-                          <li>The file is not an accepted image type.</li>
-                        )}
+                        {mapErrors(errors).map((error, i) => <li key={i}>{error}</li>)}
```
Rob--W: Is `key={i}` doing anything useful? If not, let's remove it for simplicity.

rebmullin: this is required by react. when mapping over items you need a key. do you prefer another name? or i can try using the error name..?

Rob--W: I wasn't aware of the `key` requirement (https://reactjs.org/docs/lists-and-keys.html#keys), thanks for the explanation.

Issue #849, opened 2019-10-14
Images not used in saved/set themes are not cleared from the local storage space after closing a tab session


Softvision-AndreeaCupsa:
**[Affected versions]:**
- Firefox Nightly 71.0a1, Build ID: 20191013213650
- Firefox Beta 70.0b14, Build ID:20191013213650
- Firefox Release 69.0.3, Build ID: 20191009172106

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox Color add-on zip modified in order to link to the [stage](https://color.stage.mozaws.net/) website.
- Have the previous Firefox Color add-on installed through about:debugging.

**[Steps to reproduce]:**
1. Open the profile from the prerequisites.
2. Navigate to the https://color.stage.mozaws.net/ website.
3. Click the “Custom backgrounds” section and add a valid image.
4. Click the “Random” button.
5. Close the tab.
6. Open a new tab and navigate to the https://color.stage.mozaws.net/ website.
7. Observe the local storage space.

**[Expected result]:**
- The local storage is cleared and 0MB is occupied.

**[Actual result]:**
- The local storage is not cleared and the size of the previously added image is displayed.

**[Notes]:**
- A user might easily reach the maximum limit of the local storage space if the local storage is not cleared of the added images from history after closing a tab session (images from used/saved themes must remain in storage). 
-  As a user is no longer able to navigate through history on a new session in order to delete images added in a previous session it might make it impossible to add more images.
- Due to the fact that the add-on was installed as a temporary add-on (through about:debugging) I was not able to test the scenarios including restart.
- Attached a screen recording with the issue.
![Local storage](https://user-images.githubusercontent.com/37875464/66759384-d52e2b80-eea8-11e9-85d0-700329ac2970.gif)

rebmullin:
hmm @Softvision-AndreeaCupsa, yes typically you can remove the image from local storage if its saved to a theme but I see in this case that's not done so just thinking @caitmuenster, maybe we can remove any images from local storage is there is no theme.?

caitmuenster:
@rebmullin That makes sense to me!

Softvision-AndreeaCupsa:
I have verified this issue and is no longer reproducible using the production xpi (version 2.1.5) linked to the stage server on Firefox Release 72.0.1 on Windows 10 x64, Mac 10.14.6 and Ubuntu 18.

PR #892, opened 2020-03-01
I886 adding custom bgs empty default

### LINKED ISSUE #877: First image added after randomizing a theme is not added to the theme and local storage
**[Affected versions]:**
- Firefox Nightly 72.0.1, Build ID:  20200107212822
- Firefox Beta 73.0b2, Build ID:  20200107212705
- Firefox Nightly 74.0a1, Build ID: 20200108215606

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox Color add-on zip modified in order to link to the stage website.
- Have the previous Firefox Color add-on installed through about:debugging.

**[Steps to reproduce]:**
1. Open the profile from the prerequisites.
3. Click the “Random” button.
4. Click the “Custom backgrounds” section and add a valid image.
5. Observe the custom background item section and the theme of the browser. 

**[Expected result]:**
- The elements of the added image and the occupied local storage space is correctly displayed.
- The image is displayed in the background of the current theme.

**[Actual result]:**
-  A blank space is displayed instead of the source image in the added image section.
- 0MB is occupied from the local storage space.
- The image is not displayed in the background of the current theme.

**[Notes]:**
- The issue is also reproducible after making several changes to the random theme. 
- The second image added is correctly added to the theme.
- This issue is not reproducible when adding an image after a preset theme was set. 

Image not recognized

### LINKED ISSUE #886: The "Save" button is wrongly active and actionable if a previously saved theme is clicked
**[Affected versions]:**
- Firefox Release 73.0 - Build ID: 20200207195153
- Firefox Beta 74.0b2 - Build ID: 20200211200559
- Firefox Nightly 75.0a1 - Build ID: 20200213035745

**[Affected Platforms]:**
- Windows 10 x64
- Mac 10.15.2
- Ubuntu 18.04 x64

**[Prerequisites]:**
- Have the latest version of the Firefox Color add-on modified in order to link to the [stage](https://color.stage.mozaws.net/) website.
- Have the previous Firefox Color add-on installed through "about:debugging".
	
**[Steps to reproduce]:**
3. Click the "Random" button from the top-right part of the page.
4. Click the "Save" button from the top-right part of the page.
5. Go to the "Saved themes" section and select the previously saved theme.
5. Observe the "Save" button from the top-right part of the page.

**[Expected result]:**
- The "Save" button is grayed out and unactionable.

**[Actual result]:**
- The "Save" button is active and actionable.

**[Notes]:**
- The theme is saved again and duplicated in the "Saved themes" section if the "Save" button is clicked after following the steps from above. 
- This issue is also reproducible with the 2.1.5 version of the "Firefox Color" add-on installed from the production website (https://color.firefox.com/).
- The issue is no longer reproducible after the theme is saved again.
- Attached a screen recording of the issue.
rec of the issue

rebmullin:
fixes #886 

It seems that random themes do not have a custom_backgrounds property which makes sense but in redux the present theme would show this as an empty custom_backgrounds array so it was showing as modified since these are themes are technically different. I do not believe that preset themes has this issue so I do realize there is some different implementation but imo it is better to have this default value. 

Anyway, I need to do more testing to make sure there are no regressions but I believe this fixes the issue. 

seems it also fixes #877 🎉 (but will test a bit more :)

### DIFF
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -11,7 +11,7 @@ const defaultTheme = presetThemesContext("./default.json");
 
 const IMAGE_PROPS = ["name", "tiling", "alignment"];
 
-export const themesEqual = (themeA, themeB) => {
+export const themesEqual = (themeA, themeB) => {  
   if (!!themeA !== !!themeB) {
     return false;
   }
@@ -33,17 +33,8 @@ export const themesEqual = (themeA, themeB) => {
     }
   }
 
-  if (themeA.images && !themeA.images.custom_backgrounds) {
-      themeA.images.custom_backgrounds = [];
-  }
-
-  if (themeB.images && !themeB.images.custom_backgrounds) {
-      themeB.images.custom_backgrounds = [];
-  }
-
-    const imagesA = themeA.images && themeA.images.custom_backgrounds || [];
-    const imagesB = themeB.images && themeB.images.custom_backgrounds || [] ;
-
+  const imagesA = themeA.images && themeA.images.custom_backgrounds || [];
+  const imagesB = themeB.images && themeB.images.custom_backgrounds || [] ;
     if (imagesA.length !== imagesB.length) {
       return false;
     }
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -40,17 +40,10 @@ export const themesEqual = (themeA, themeB) => {
   if (themeB.images && !themeB.images.custom_backgrounds) {
       themeB.images.custom_backgrounds = [];
   }
-  
-  const hasCustomImagesA =
-    "images" in themeA && "custom_backgrounds" in themeA.images;
-  const hasCustomImagesB =
-    "images" in themeB && "custom_backgrounds" in themeB.images;
-  if (hasCustomImagesA !== hasCustomImagesB) {
-    return false;
-  }
-  if (hasCustomImagesA && hasCustomImagesB) {
-    const imagesA = themeA.images.custom_backgrounds;
-    const imagesB = themeB.images.custom_backgrounds;
+
+    const imagesA = themeA.images && themeA.images.custom_backgrounds || [];
+    const imagesB = themeB.images && themeB.images.custom_backgrounds || [] ;
+
     if (imagesA.length !== imagesB.length) {
       return false;
     }
@@ -64,7 +57,6 @@ export const themesEqual = (themeA, themeB) => {
         }
       }
     }
-  }
 
   // TODO: Skipping title equality, because user themes don't have titles yet.
 
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -33,6 +33,14 @@ export const themesEqual = (themeA, themeB) => {
     }
   }
 
+  if (themeA.images && !themeA.images.custom_backgrounds) {
+      themeA.images.custom_backgrounds = [];
+  }
+
+  if (themeB.images && !themeB.images.custom_backgrounds) {
+      themeB.images.custom_backgrounds = [];
+  }
+  
   const hasCustomImagesA =
     "images" in themeA && "custom_backgrounds" in themeA.images;
   const hasCustomImagesB =
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -35,19 +35,19 @@ export const themesEqual = (themeA, themeB) => {
 
   const imagesA = themeA.images && themeA.images.custom_backgrounds || [];
   const imagesB = themeB.images && themeB.images.custom_backgrounds || [] ;
-    if (imagesA.length !== imagesB.length) {
-      return false;
-    }
-    for (let idx = 0; idx < imagesA.length; idx++) {
-      for (let propIdx = 0; propIdx < IMAGE_PROPS.length; propIdx++) {
-        if (
-          imagesA[idx][IMAGE_PROPS[propIdx]] !==
-          imagesB[idx][IMAGE_PROPS[propIdx]]
-        ) {
-          return false;
-        }
+  if (imagesA.length !== imagesB.length) {
+    return false;
+  }
+  for (let idx = 0; idx < imagesA.length; idx++) {
+    for (let propIdx = 0; propIdx < IMAGE_PROPS.length; propIdx++) {
+      if (
+        imagesA[idx][IMAGE_PROPS[propIdx]] !==
+        imagesB[idx][IMAGE_PROPS[propIdx]]
+      ) {
+        return false;
       }
     }
+  }
 
   // TODO: Skipping title equality, because user themes don't have titles yet.
 
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -11,7 +11,7 @@ const defaultTheme = presetThemesContext("./default.json");
 
 const IMAGE_PROPS = ["name", "tiling", "alignment"];
 
-export const themesEqual = (themeA, themeB) => {  
+export const themesEqual = (themeA, themeB) => {
   if (!!themeA !== !!themeB) {
     return false;
   }
--- a/src/web/lib/components/AppHeader/index.js
+++ b/src/web/lib/components/AppHeader/index.js
@@ -43,7 +43,7 @@ export const AppHeader = props => {
     if (shouldUpdate) {
       syncImages();
     }
-  }, [props.themeCustomBackgrounds]);
+  }, [shouldUpdate]);
 
 
   const handleExportClick = () => {
@@ -84,6 +84,7 @@ export const AppHeader = props => {
     if (toDelete.length) {
       props.deleteImages(toDelete);
     }
+    setUpdate(false);
   };
 
   const headerButton = (
--- a/src/preset-themes/default.json
+++ b/src/preset-themes/default.json
@@ -50,6 +50,9 @@
     }
   },
   "images": {
-    "additional_backgrounds": ["./bg-009.svg"]
+    "additional_backgrounds": [
+      "./bg-009.svg"
+    ],
+    "custom_backgrounds": []
   }
 }
--- a/src/preset-themes/default.json
+++ b/src/preset-themes/default.json
@@ -52,7 +52,6 @@
   "images": {
     "additional_backgrounds": [
       "./bg-009.svg"
-    ],
-    "custom_backgrounds": []
+    ]
   }
 }
--- a/src/preset-themes/default.json
+++ b/src/preset-themes/default.json
@@ -50,8 +50,6 @@
     }
   },
   "images": {
-    "additional_backgrounds": [
-      "./bg-009.svg"
-    ]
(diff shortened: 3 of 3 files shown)


rebmullin:
hey @Rob--W, I have done some testing and everything seems okay.  I wanted to know your thoughts about this?

rebmullin:
hey @Rob--W ,thanks so much for the feedback: )

I am still on the fence that updating this file is better tho :/. why not just set it as this value as the default in the set up so we don't have to reset it to this anyway? Anyway, I ended up reverting the changes to the default file and changed the file you noted instead. However, I am not exactly sure if the changes is what you had in mind so please let me know. Basically I just set these themes to have an empty array if nothing exists and removed these checks if hasCustomImages here since it's will always be true after setting up the default array. in my prelim testing it still seems to work for both issues noted.

Rob--W:
> why not just set it as this value as the default in the set up so we don't have to reset it to this anyway?

The root cause of the bug is the fact that `themesEqual` fails at treating two equal themes as equal. Updating the theme template is just a work-around, anything else that relies on `themesEqual` would still suffer from this bug. Since it is not difficult to fix `themesEqual`, I would rather see the method fixed than the bug being worked around.

> Basically I just set these themes to have an empty array if nothing exists

I would not expect a method with the name `themesEqual` to mutate the input parameter. This side effect may have unexpected results. A few lines down, you already fall back to `|| []`, so it is not even necessary to assign to `theme[AB].images.custom_backgrounds`.

> and removed these checks if hasCustomImages here since it's will always be true after setting up the default array. in my prelim testing it still seems to work for both issues noted.

Looks good. Please fix up the indentation though. And revert the unnecessary whitespace changes in the theme template.

rebmullin:
> I would not expect a method with the name themesEqual to mutate the input parameter. This side effect may have unexpected results. A few lines down, you already fall back to || [], so it is not even necessary to assign to theme[AB].images.custom_backgrounds.

hmm I get that maybe mutating these themes is not ideal but removing this will keep the bug. The comparison is comparing an empty array with no custom_bgs somewhere else (I believe here: modifiedSinceSave). So I'll need to dig into this more if this needs to be reverted..

Rob--W:
If you are not able to find why the bug continues to happen without mutating the input, let's then just add `custom_backgrounds` to the default JSON. I originally asked for the `themesEqual` method to be fixed, in case a saved theme didn't have the `custom_backgrounds`.

(in case it itsn't clear: make the change to the default template AND fix `themesEqual` but without mutating the input.)

rebmullin:
@Rob--W, before I update default.json, I did find where present themes was being set to default to an empty array when there were no custom images. If I update this and check for this in various places the following changes seem to work: 

```diff
diff --git a/src/lib/store.js b/src/lib/store.js
index d6f3a22..5034a53 100755
--- a/src/lib/store.js
+++ b/src/lib/store.js
@@ -99,7 +99,7 @@ export const selectors = {
   themeCanRedo: state => state.theme.future.length > 0,
   themeCustomImages: state => state.images.images,
   themeCustomBackgrounds: state =>
-    state.theme.present.images.custom_backgrounds || [],
+    state.theme.present.images.custom_backgrounds,
   themeHasCustomBackgrounds: state => {
     const backgrounds = selectors.themeCustomBackgrounds(state);
     return !!backgrounds && backgrounds.length > 0;
diff --git a/src/lib/themes.js b/src/lib/themes.js
index f209a52..abf4117 100644
--- a/src/lib/themes.js
+++ b/src/lib/themes.js
@@ -33,14 +33,6 @@ export const themesEqual = (themeA, themeB) => {
     }
   }
 
-  if (themeA.images && !themeA.images.custom_backgrounds) {
-      themeA.images.custom_backgrounds = [];
-  }
-
-  if (themeB.images && !themeB.images.custom_backgrounds) {
-      themeB.images.custom_backgrounds = [];
-  }
-
     const imagesA = themeA.images && themeA.images.custom_backgrounds || [];
     const imagesB = themeB.images && themeB.images.custom_backgrounds || [] ;
 
diff --git a/src/web/lib/components/AppHeader/index.js b/src/web/lib/components/AppHeader/index.js
index 846e5dc..d482a07 100644
--- a/src/web/lib/components/AppHeader/index.js
+++ b/src/web/lib/components/AppHeader/index.js
@@ -58,7 +58,7 @@ export const AppHeader = props => {
   const syncImages = () => {
     let currentImages = new Set();
 
-    props.themeCustomBackgrounds
+    (props.themeCustomBackgrounds || [])
       .forEach(({ name }) => currentImages.add(name));
 
     let savedImagesInThemes = new Set();
diff --git a/src/web/lib/components/ThemeCustomBackgroundPicker/index.js b/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
index 4e39b58..7505ed3 100755
--- a/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
+++ b/src/web/lib/components/ThemeCustomBackgroundPicker/index.js
@@ -25,11 +25,12 @@ export class ThemeCustomBackgroundPicker extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { themeCustomBackgrounds, themeCustomImages } = this.props;
+    const { themeCustomBackgrounds = [], themeCustomImages } = this.props;
+    const { themeCustomBackgrounds: prevThemeCustomBackgrounds = []} = prevProps;
 
     // sync images in local storage with custom backgrounds in preview on undo/redo
     // when custom background changes.
-    if (prevProps.themeCustomBackgrounds.length !== themeCustomBackgrounds.length) {
+    if ( prevThemeCustomBackgrounds.length !== themeCustomBackgrounds.length) {
       const keys = Object.keys(themeCustomImages);
 
       themeCustomBackgrounds.forEach(({ name }) => {
@@ -94,7 +95,7 @@ export class ThemeCustomBackgroundPicker extends React.Component {
       addImage,
       updateImage,
       themeHasCustomBackgrounds,
-      themeCustomBackgrounds,
+      themeCustomBackgrounds = [],
       storageErrorMessage
     } = this.props;
     const label = themeHasCustomBackgrounds
@@ -179,7 +180,7 @@ const BackgroundList = SortableContainer(props => {
   const {
     clearCustomBackground,
     updateCustomBackground,
-    themeCustomBackgrounds,
+    themeCustomBackgrounds = [],
     themeCustomImages
   } = props;

```

if this is too hard to read, I can push it up...

Anyway, I did some prelim testing and the above diff seems to work too. I wanted to run it by you first before updating default.json seems it seemed like you wanted to do that as a last resort. (note does not keep the mutation in themesEqual but it does remove the hasCustomImage checks there etc). 

please let me know what you think :)

Rob--W:
The required changes after the removal of `|| []` from `themeCustomBackgrounds` results in less readable code IMO. And it would probably also cause errors in locations where `themeCustomBackgrounds` is expected to be an array but not updated to handle the case of not being set, like here: https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L106-L109 and https://github.com/mozilla/FirefoxColor/blob/31019674c050dd46ccaf6889108a408e1fa55328/src/web/lib/middleware.js#L79-L84

How would the proposed diff fix this bug?

rebmullin:
>The required changes after the removal of || [] from themeCustomBackgrounds results in less readable code IMO. And it would probably also cause errors in locations where themeCustomBackgrounds is expected to be an array ...

hmm It looks like it's still an array but I am not sure why without further investigation...

But if it's less readable I am happy to return it to just updating the default.json.

Rob--W:
> But if it's less readable I am happy to return it to just updating the default.json.

The minimum to approve this PR is that it fixes the issue in a reasonable way. Updating default.json does look like a reasonable approach, especially when combined with a fix in `themesEqual`.

I did expect the update of `themesEqual` to have resolved the problem on its own. I would like to understand why it was not sufficient, but if you don't have time to find that out, I would be fine with approving the PR with the reasonable fix.

rebmullin:
hey @Rob--W,

RE #886 - well we know the themes aren't coming up equal b/c in one theme it’s set to an empty array and in other the property doesn’t exist. After some digging, I believe I finally found the culprit:

https://github.com/mozilla/FirefoxColor/blob/master/src/web/lib/components/PaginatedThemeSelector/index.js#L45

I try can modifying this so it only does this when there is an actual custom background.

We can also adjust the themesEaual as we did in the current PR. I believe both / either fix this issue. For now I am just modifying themesEqual here..

RE #877 - based on some prelim testing, it seems there was bug in sync images. So I was able to fix it by updating useEffect here.

I am going to keep these combined if thats okay and push these changes up. If you disagree w/these then I am going to update the default.json.

rebmullin:
I know you gave me the go ahead already on the default.json but I wanted to find out what was going on here. 

I still think that adding this default value to default.json makes the most sense tho so we don't have to adjust it in other places, like I am doing here.

But of course I'll defer to you and I am definitely curious to know what you think :)

src/web/lib/components/AppHeader/index.js
```diff
@@ -43,7 +43,7 @@ export const AppHeader = props => {
     if (shouldUpdate) {
       syncImages();
     }
-  }, [props.themeCustomBackgrounds]);
+  }, [shouldUpdate]);
```
Rob--W: This change is correct iff the module only needs to account for changes initiated through the UI, not for external changes (such as changes in storage, e.g. by changing themes in a different tab).

Can you confirm whether this is your intended change?

src/lib/themes.js
```diff
@@ -33,16 +33,8 @@ export const themesEqual = (themeA, themeB) => {
...
   }
 
-  const hasCustomImagesA =
-    "images" in themeA && "custom_backgrounds" in themeA.images;
-  const hasCustomImagesB =
-    "images" in themeB && "custom_backgrounds" in themeB.images;
-  if (hasCustomImagesA !== hasCustomImagesB) {
-    return false;
-  }
-  if (hasCustomImagesA && hasCustomImagesB) {
```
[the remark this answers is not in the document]
rebmullin: yeh something's going on with my prettier. i'll look into that..

src/web/lib/components/AppHeader/index.js
rebmullin: hmm.. it seems that either way, adding a custom image in another tab will cause a similar issue (#877).  In another issue #849, we decided to clear out the image from local storage if the user refreshed and the image wasn't saved so not too sure how to get around this. I can check for the image before displaying it in the background list but regardless it seems that it's irrelevant to this change here.

rebmullin: @Rob--W, I am wondering if we can merge this in as is (or update the default.json file). Either way seems to fix the issue at hand within the same tab but if you open in another tab, it's still a problem. To me this seems less common UI (?) and since it currently does this - so it's not something that this PR is breaking - I am wondering if we can have a follow up issue to look into another solution (keeping in mind #849). 

What do you think?

PR #899, opened 2020-03-17
Fixes #878: Notification message

### LINKED ISSUE #878: Notification message does not point to “Share” button at click
[Affected versions]:
Firefox Nightly 74.0a1 

[Affected Platforms]:
All Windows 
All Mac 
All Linux 

[Prerequisites]:
Have a new Firefox profile with the latest version of the "Firefox Color" add-on  

[Steps to reproduce]:
2. Navigate to color.firefox.com   and install Firefox Color add-on
3. Click on “Share” button from top right side of the page 
4. Observe the notification message  

[Expected result]:
Notification message points to “Share” button.

[Actual result]:
Notification message points to “Random” button.

[Notes]:
This issue is not reproducible before "Firefox Color" add-on installation. 
Attaching a screenshot with the issue.
share-button

Je-ni:
image

This is the fixed preview. #878

### DIFF
--- a/src/web/lib/components/ThemeUrl/index.scss
+++ b/src/web/lib/components/ThemeUrl/index.scss
@@ -17,7 +17,7 @@
   }
 
   &.extension {
-    right: 182px;
+    right: 70px;
   }
 
   &::after {


Je-ni:
What are magic numbers? And how can we enhance the project to minimize its use? @Rob--W

Rob--W:
A [magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) is a numeric value in source code from which it is not clear why the value is as it is. In this case, the number is a value in a style sheet, which you probably found by trial-and-error, so there is probably no deeper meaning behind it.

To show how it could have been done otherwise, I'll give an example inspired by the screenshot here. Note that it may not reflect reality, it is just an example.

The interface has the following appearance, and the goal is to align the left side of the `notification` panel with the middle of the Share button.

```
[Save][Share][Export][padding]
 ^
 [notification]
```

You fixed it by replacing a number with another number. If the Export button or padding ever becomes wider, then the notification would be off again.

Now, to avoid that bug, if there were constants/variables for the horizontal distances of the Share/Export/padding blocks, then it would be easier to maintain the goal. And that is in fact possible, even with plain CSS using [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Variables) and the [`calc` CSS function](https://developer.mozilla.org/en-US/docs/Web/CSS/calc). Now, before you go and submit pull requests for everything, do note that when used sparingly and responsibly, the use of variables and calc can improve readability and maintainability. However, excessive use of it (e.g. encoding far-fetched relations between CSS property values) harms maintainability.

With that being said, here is a demonstration of how the use of CSS variables can make it easier to explicitly encode the relation between different parts of a stylesheet:

(see https://jsfiddle.net/47jyh562/ for a full example with colors for easier visuals)

```html
<header>
  <div class="but">Save</div>
  <div class="but">Share</div>
  <div class="but">Export</div>
  <div class="notification">Notification</div>
</header>
```
```css
:root {
  --button-width: 150px;
  --top-bar-right-padding: 30px;
  --notification-width: 80px;
}

.notification {
  position: absolute;
  right: calc(var(--button-width) * 1.5 + var(--top-bar-right-padding) - var(--notification-width));
  width: var(--notification-width);
}
header {
  padding-right: var(--top-bar-right-padding);
  position: relative;
  display: inline-block;
}
.but {
  width: var(--button-width);
  display: inline-block;
}
```

Je-ni:
Okay. I get what you mean by magic numbers. It'll require some refactoring to remove and avoid future CSS magic numbers.

Rob--W:
> It'll require some refactoring to remove and avoid future CSS magic numbers.

Yep. And for the specific case here, it's not worth the effort considering that it may negatively impact the readability and maintainability of the affected code.

Issue #866, opened 2019-11-12
Shared Theme Link overrides present theme by default theme.


ankiiitraj:
If you open link of shared theme which you are already using, the present theme is replaced by some other default theme.

Steps to reproduce:
- Open color.firefox.com and click on share then copy the link.
- Open the copied link 

![Peek 2019-11-12 21-32](https://user-images.githubusercontent.com/48787278/68690516-26a90380-0598-11ea-9fea-88e9f38b3071.gif)

ankiiitraj:
I want to solve this issue, Can anyone please provide me some hints.

caitmuenster:
Thanks @ankiiitraj, this one is interesting. @Rob--W, is this issue appropriate for a `good-first-bug` label?

anuragakella:
Hey, I'd like to contribute to this. Can someone guide through?

ankiiitraj:
Hi I'm not current working on this you can take on. Thanks

anuragakella:
> 
> 
> Hi I'm not current working on this you can take on. Thanks

Hey, cool I'd love to work on it, but can someone guide me? @Rob--W, @caitmuenster ..

Rob--W:
This was not marked as good first bug, but if you're willing to debug the issue I'm willing to mentor it. The first step is to set up a development environment (the README shows how to do that). After that you can dive into this bug.

The web page is designed to work with and without extension. To communicate between the web page and the extension, the `window.postMessage` API is used. When the extension is not installed, the web page will work and keep all data locally. When the extension *is* installed, then the extension will receive messages from the page (e.g. setTheme, fetchTheme, ping) and also reply (e.g. fetchedTheme, pong).

I was not able to consistently reproduce this bug, but when it occurred, the "setTheme" message was sent by the web page before "fetchTheme":

- "setTheme" (in response to "pong" from the extension, which is in response to "ping" from the web page) - https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L157
- "fetchTheme" - https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L300

Obviously, when the web page sends a `setTheme` message before the extension had a chance to respond to `fetchTheme`, then the web page does not know about the browser's actual theme, and the extension will receive the default theme (and overwrite the extension's actual theme with that default theme).

[...]

PR #910, opened 2020-03-26
Bug Fix: Shared Theme Link overrides present theme by default theme. #866

anuragakella:
Fixed a bug that resets the theme to a default theme when a shared theme link is loaded.

For the code here:
https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L153-L157

First, I followed your suggestion and after a few changes I added an if condition:
Now the same lines look like:
```
if (!params.theme) {
  const state = store.getState();
  postMessage("addImages", {
    images: selectors.themeCustomImages(state)
  });
  postMessage("setTheme", { theme: selectors.theme(state) });
}
```
Since the bug occurs only when a shared link is opened, I decided to add a flag and check if the current page was a shared link or not using `params.theme` to check for ?theme, and it works!
if the page IS a shared link, then I added another `postMessage("fetchTheme");` here at line 281, the first line in the else block:
https://github.com/mozilla/FirefoxColor/blob/a43d5c66c7dc6f644f4e42e6a53ed1df9e19f099/src/web/index.js#L280-L290

I did that because, if the theme was not a shared link, the existing code already did a fetchTheme and worked properly. But if `params.theme` exists, there was no fetchTheme call.

Thanks!


src/web/index.js
```diff
@@ -278,6 +281,8 @@ if (!params.theme) {
   // The add-on may never answer, so start the loader delay.
   startLoaderDelay();
 } else {
+  // trigger a message to fetch and apply the current theme that the browser already has.
+  postMessage("fetchTheme");
```
Rob--W: There is no guarantee that the extension is listening at this point, nor that the extension will reply with "fetchedTheme" before the hello/pong message. As a result, the fix here is no reliable.

In #866, I suggested to add a new flag (to the `setTheme` and possibly `addImages` call), so that the extension's background.js can choose to ignore the theme assignment if the user has already set a theme before. Have you tried that?

anuragakella: Oh, I didn't quite understand that the first time. So by adding a flag you meant something like:
`postMessage("setTheme", { theme: theme, userHasTheme: true/false })` 
and then I do a condition to check that flag inside `messageHandlers` in `background.js` before applying the theme?
Is that what you meant?
Sorry and thanks!

Rob--W: Yes exactly. And I would call the flag something like `defaultFromWeb: true`, and add a comment of why this was done (maybe based on the description that I wrote in the original issue).

Issue #845, opened 2019-10-10
Create a static page so we can download dev and stage's xpi


rebmullin:
We should create a static page which links out from the "Get Firefox Color" btn on dev and stage only. This page will have links so that we can download the dev and stage xpis.

The I841 [fix](#842) only addressed the url for prod so this is a follow up for that. See more here:

https://github.com/mozilla/FirefoxColor/pull/842#discussion_r332041803

Rob--W:
Were you planning to work on this @rebmullin ? To test the current master, I wanted to push it to the `development` branch to update gh-pages, but that did not work because in the current build process, there is an attempt to sign the add-on: https://github.com/mozilla/FirefoxColor/blob/cb5a6437739569629b315b895e6c0f2f1d612a3c/package.json#L60

The fix is to not sign the add-on, but to build an unsigned addon (e.g. through a `xpi:dev` command similar to `xpi:stage` / `xpi:production`) and link to that via the button, or the page that is described here. After removing the `sign` step, it will be possible to preview changes by pushing to the development branch.

rebmullin:
hey @Rob--W,

yes I still plan to work on this :) Sorry for the delay and that this is blocking the dev deploy. I posted a PR - #912 - which I am hoping can be an interim fix 🤞 

Regarding this static page. I have a few questions:

1. you suggested linking out to the build urls but I am not sure - what would be the URLS for each env. I guess that's not too clear to me. 
2. where would this page live?

Rob--W:
> 1. you suggested linking out to the build urls but I am not sure - what would be the URLS for each env. I guess that's not too clear to me.

For stage and prod, the URLs are here: https://github.com/mozilla/FirefoxColor#build-test-and-publish-add-on
These names are defined here:
https://github.com/mozilla/FirefoxColor/blob/c55fc2cad3ebe8591433082a2a255141f21c170c/package.json#L45-L46

There is no development name yet, but I was thinking of "firefox-color-dev-unsigned.xpi"

> 2. where would this page live?

Something like `build/web/testing.html` (from `src/web/`?). It would definitely be used as SITE_URL on the development branch, and possibly also on the stage branch. Not on production though.

Issue #907, opened 2020-03-25
Incorrect SITE_URL value is used when the website is built


Rob--W:
STR:

1. View the source of https://color.firefox.com/ or https://color.stage.mozaws.net/

Expected:
- Any reference to external resources should point to the site itself, i.e. color.firefox.com or color.stage.mozaws.net.

Actual:
- `localhost:8080` is used instead:

```
    <meta property="og:url" content="http://localhost:8080/" />
    <meta property="og:type" content="website" />
    <meta property="og:image" content="http://localhost:8080/images/color-fb.jpg" />
```

The site is built via the `npm run build` call at https://github.com/mozilla/FirefoxColor/blob/cb5a6437739569629b315b895e6c0f2f1d612a3c/.circleci/config.yml#L23

which is implemented by 
https://github.com/mozilla/FirefoxColor/blob/cb5a6437739569629b315b895e6c0f2f1d612a3c/package.json#L37-L38

The config is at https://github.com/mozilla/FirefoxColor/blob/cb5a6437739569629b315b895e6c0f2f1d612a3c/webpack.web.js#L40-L44
and read from the `SITE_URL` environment variable at https://github.com/mozilla/FirefoxColor/blob/cb5a6437739569629b315b895e6c0f2f1d612a3c/webpack.common.js#L13-L15

To fix this issue, the `SITE_URL` environment variable should be assigned a value in `config.yml`. For more info, see https://circleci.com/docs/2.0/env-vars/

Rob--W:
CircleCI does not seem to support branch-specific environment variables, according to the resources that I have consulted:

- https://ideas.circleci.com/ideas/CCI-I-302
- https://stackoverflow.com/questions/44749061/circleci-different-value-to-environment-variable-according-to-the-branch
- (and a few others)

I think that we have to add some inline shell scripting (or a shell script to be sourced) to the circleci config, to dynamically set the environment variable from there.

PR #912, opened 2020-03-27
potential fix for dev deploys / static test page

rebmullin:
hey @Rob--W,

I am still plan on working on #845 (I'll post my questions for that over there) but in the interim please let me know if this will help.

I am wasn't 100% sure about updating the webpack.common file tho. I believe the config.yml will override the value there so if that's the case (which we want for prod), then this should be okay.

NOTE: this reverts some of what is added here: https://github.com/mozilla/FirefoxColor/pull/842


package.json
```diff
@@ -58,7 +58,7 @@
     "test:js": "mocha --require src/lib/test-setup.js --recursive \"src/**/*-test.js\"",
     "deploy": "gh-pages -x -d build/web -r \"https://$GH_TOKEN@github.com/mozilla/FirefoxColor.git\"",
     "release:base": "npm-run-all clean build:web sign && mv addon.xpi build/web && npm run deploy",
-    "release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/firefox/addon/firefox-color/' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
```
Rob--W: This URL is closer to production than a direct link to the xpi, but neither of the options work because the xpi needs to be signed, and CircleCI does currently not have valid credentials to generate a signed xpi file. I thought of removing the signing step, in favor of linking to unsigned files + clear instructions on how to load it (#845). But that requires #845 to be resolved first.

rebmullin:
hey @Rob--W, 

I decided to add the work for the static page here since it's all related.  

I am not 100% the new dev script is correct. I believe I have it doing the same same steps as stage/prod to create an unsigned addon - can you please confirm?

src/web/testing.html
```diff
@@ -0,0 +1,27 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width" />
+    <title>Unsigned Addons</title>
+  </head>
+  <body>
+    <ul>
+      <li>
+        <a
+          href="https://mozilla.github.io/FirefoxColor/build/web/firefox-color-dev-unsigned.xpi"
```
Rob--W: Use a relative URL so that the links point to the xpis that are built with the current deployment.

E.g.

```html
<li><a href="firefox-color-dev-unsigned.xpi">Unsigned add-on for localhost:8080</a>
<li><a href="firefox-color-stage-unsigned.xpi">Unsigned add-on for color.stage.mozaws.net</a>
<li><a href="firefox-color-unsigned.xpi">Unsigned add-on for color.firefox.com</a>
```

package.json
```diff
@@ -58,7 +58,7 @@
...
+    "release:dev": "cross-env SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run package && mv addon.xpi build/web/firefox-color-dev-unsigned.xpi npm run deploy"
```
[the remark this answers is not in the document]
rebmullin: hmm i actually tried that initially - xpi:dev - but this was posing problems with WEB_EXT_API_KEY and nset WEB_EXT_API_SECRET. 
So I'll have to circle back to this..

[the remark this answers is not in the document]
Rob--W: When run from CI, those environment variables are explicitly cleared: https://github.com/mozilla/FirefoxColor/blob/fda866d2a3ee506261925db287856426cf4d80d2/.circleci/config.yml#L25-L27

I guess that you called something like `xpi:dev` from `release:dev`? That is not needed, all build steps under `build_and_test` are run in CI, see the .circleci/config.yml file.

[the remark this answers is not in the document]
rebmullin: > I guess that you called something like xpi:dev from release:dev

yep : ) this is what I did. thanks

src/web/index.js
```diff
@@ -28,7 +28,9 @@ log("startup");
 
 const clipboard = new Clipboard(".clipboardButton");
 
-const addonUrl = process.env.ADDON_URL;
+const siteUrl = process.env.SITE_URL;
+const addonUrl = siteUrl.includes("color.firefox.com") ? process.env.ADDON_URL : `${siteUrl}build/web/testing.html`;
```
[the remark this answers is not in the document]
rebmullin: So I've reverted this in an updated the config for dev. 

Should I be doing a similar thing for stage here?

package.json
```diff
@@ -58,7 +58,7 @@
     "test:js": "mocha --require src/lib/test-setup.js --recursive \"src/**/*-test.js\"",
     "deploy": "gh-pages -x -d build/web -r \"https://$GH_TOKEN@github.com/mozilla/FirefoxColor.git\"",
     "release:base": "npm-run-all clean build:web sign && mv addon.xpi build/web && npm run deploy",
-    "release:dev": "cross-env ADDON_URL='https://addons-dev.allizom.org/firefox/addon/firefox-color/' SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run release:base"
+    "xpi:dev": "cross-env SITE_URL='https://mozilla.github.io/FirefoxColor/' SITE_ID=github npm run package && mv addon.xpi build/web/firefox-color-dev-unsigned.xpi"
```
Rob--W, line 61: Could you move this to the line before `xpi:stage`? A bit easier to maintain.

The `release:dev` line here can be restored (or remove it altogether with `release:base`, since both are not used any more). The README still mentions `release:dev`, which is not accurate any more. I can update the documentation if you aren't sure on what to put there.

src/web/testing.html
```diff
@@ -0,0 +1,28 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width" />
+    <title>Unsigned Addons</title>
+  </head>
+  <body>
+    <ul>
+      <li>
+        <a href="firefox-color-dev-unsigned.xpi">
+          Unsigned add-on for mozilla.github.io/FirefoxColor</a
```
Rob--W, line 12: `</a` looks a bit odd. Could you move it to the next line so there is `</a>`?

.circleci/config.yml
```diff
@@ -40,13 +40,15 @@ jobs:
   deploy_development:
     docker:
       - image: circleci/node:10.18.1@sha256:8230f70c03b6ecf521a34125fb090ec4b47068205454e5820859c00260c9bd23
+    environment:
+      ADDON_URL: testing.html
```
Rob--W, line 44: This doesn't work because the site is built in the `test_and_build` step, using a single config for all. The proper fix for that is will be part of https://github.com/mozilla/FirefoxColor/issues/907

The testing.html page is a useful overview of available xpis for testing, so I don't mind that the Install button is not pointing to the desired page.

rebmullin:
hey @Rob--W, I followed some links noted in the other issue. I'm still not 100% that I have this right, so please let me know.

.circleci/config.yml
```diff
@@ -19,6 +17,7 @@ jobs:
       - run: greenkeeper-lockfile-update
       - run: npm ci
       - run: greenkeeper-lockfile-upload
+      - run: bash ./compile.sh
```
Rob--W, line 20: This does not work as expected, because running `bash` creates a new subshell with its own environment variables. The use of `export xxx=...` will change the xxx environment variable as long as that script runs, but won't affect the environment variables outside of it.

(in general)
To get the desired behavior, the shell script should be sourced. Usually something like `source compile.sh` instead of `bash compile.sh`. Then the commands from the file will be read and executed in the current subshell.

(in the special case of CircleCI)
According to CircleCI's documentation, it sources `$BASH_ENV` before each step (Ctrl-F `$BASH_ENV` in https://circleci.com/docs/2.0/env-vars/ ). You can take advantage of this by prepending a new step (right after `steps`) that generates the environment variable, as follows:

```yaml
- run: bash generate-env.sh >> $BASH_ENV
```

`generate-env.sh` could be based on what you have right now, except `exit 0` being replaced with the following, and `export` removed from your original code, because it is used as a shell variable instead of an environment variable:

```sh
...
elif [ "${CIRCLE_BRANCH}" == "production" ]
then
  SITE_URL="https://color.firefox.com/"
  ADDON_URL="https://addons.mozilla.org/firefox/addon/firefox-color/"
else
  echo "Unknown branch: ${CIRCLE_BRANCH}" >&2
  exit 1
fi

# This will be appended to $BASH_ENV in .circleci/config.yml
cat <<ENV_HERE
export SITE_URL="${SITE_URL}"
export ADDON_URL="${ADDON_URL}"
ENV_HERE
```

README.md
```diff
@@ -55,19 +55,9 @@ that are handy to know about:
 Deploying a development release consists of pushing to the `development` branch
 on this repo. Production release consists of pushing to the `production` branch.
 
-The script `npm run release:dev` in `package.json` takes care of the following:
+The script `npm run deploy` in `package.json` takes care of the following:
```
Rob--W, line 58: Use this (copyedit if needed):

```
Upon push, CircleCI will run the following steps, as defined in the `.circleci/config.yml` file:

* Run gen-environment.sh to define the `SITE_URL` and `ADDON_URL` applicable to the current branch.

* Run code linter

* Build the site for the current branch

* Build the add-ons for all build targets (development, stage, release).

* Run tests on the current branch.

When pushed to the development branch, `npm run deploy` is run to deploy the site to Github Pages.
```

README.md
```diff
@@ -104,7 +95,7 @@ Every release requires a version bump, because version numbers cannot be reused.
   - The user clicks on the "Install" button and after granting permissions, a new tab opens to the addon's home page.
 
 * Coming from the addon's home page:
-  - The user can click on the "Get Firefox Color" button which will link the user back the AMO site in the respective environment.
+  - The user can click on the "Get Firefox Color" button which will link over to the AMO site when coming from production or it will link back to the testing.html. The testing page displays links to the unsigned.xpis for each environment.
```
Rob--W, line 98: This is a bit too detailed. Let's summarize, to something like this:

> The user can click on the "Get Firefox Color" button which will direct the user to a page from where the add-on can be installed, usually AMO.

rebmullin:
> elif [ "${CIRCLE_BRANCH}" == "production" ]
then
  SITE_URL="https://color.firefox.com/"
  ADDON_URL="https://addons.mozilla.org/firefox/addon/firefox-color/"
else
  echo "Unknown branch: ${CIRCLE_BRANCH}" >&2
  exit 1
fi

If I use the "else" as suggested above ^, it hits the "Unknown branch..." error here:



It seems that the branch is the pull request branch so it fails..


if we need a default /else maybe we can just do something like ~

```bash
...
elif [ "${CIRCLE_BRANCH}" == "stage" ]
then
  SITE_URL="https://color.stage.mozaws.net/"
  ADDON_URL="testing.html"
else
  SITE_URL="https://color.firefox.com/"
  ADDON_URL="https://addons.mozilla.org/firefox/addon/firefox-color/"
fi
```

Rob--W:
> If I use the "else" as suggested above ^, it hits the "Unknown branch..." error here:
> 
> 
> 
> It seems that the branch is the pull request branch so it fails..

That is somewhat expected; the current branch is `pull/912` (and the target branch `master`, but I don't know if there is a variable for it).

When built from a pull request, the value of `SITE_URL` and `ADDON_URL` are not really important. To have a sensible default, let's not set any environment variables at all for that case, and print the current branch like we do right now (+ the reason for not setting variables).

Note: in the current PR you are not exporting the variables like I requested. For it to work correctly, you must include the `cat <<...` that I showed in my snippet. Please follow the format of my previous comment (without the `exit 0`).

rebmullin:
> To have a sensible default, let's not set any environment variables at all for that case, and print the current branch like we do right now (+ the reason for not setting variables).

okay so I left the "Unknown branch.." without the error exit code so it doesn't fail in circle ci.

> Note: in the current PR you are not exporting the variables like I requested. For it to work correctly, you must include the cat <<... that I showed in my snippet. Please follow the format of my previous comment (without the exit 0).

woops! yep i missed that. 

How's it look now? : )

rebmullin:
thanks @Rob--W for all the help with this one!

PR #917, opened 2020-04-05
pkg lockfile is outdated - update

rebmullin:


rebmullin:
tested and looks good to merge

PR #925, opened 2020-04-25
updating readme

rebmullin:
just adding some extra notes for how to load the extension locally. let me know if the formatting here is a problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,28 @@
    This will start a webpack-dev-server instance at port 8080 and start a
    watcher that will rebuild the browser extension with every file change.
 
-1. In Firefox 57 + open `about:debugging` and load the
-   `build/extension/manifest.json` file.
+1. To load the extension locally you can do this in the following ways:
 
-1. Visit `http://localhost:8080` to see the web-based theme editor - changes
+	  - In Firefox 57 + open `about:debugging` and load the
+	      `build/extension/manifest.json` file.
+
+	  - run the following steps:
+
+	  	```
+	    	rm -rf addon.xpi
+	    	npm run package
+	    	npm run start
+	   	```
+
+		  This will add an xpi file to the root of your project which you can drag to the browser. You will need to give it permissions when done with this latter method.
+
+
+		To debug the background page, you can go to Tools -> Web Developer -> Remote Debugging and click on the "Inspect" button.
+
+		To toggle the addon on and off or remove, you can go to `about:addons`
+
+
+4. Visit `http://localhost:8080` to see the web-based theme editor - changes
    should be relayed through the temporarily installed add-on and alter the
    browser theme
 
--- a/README.md
+++ b/README.md
@@ -19,25 +19,29 @@
    This will start a webpack-dev-server instance at port 8080 and start a
    watcher that will rebuild the browser extension with every file change.
 
-1. To load the extension locally you can do this in the following ways:
+1. To activate the extension:
 
-	  - In Firefox 57 + open `about:debugging` and load the
-	      `build/extension/manifest.json` file.
+   1. Use Firefox 57 + Dev or Nightly browser
 
-	  - run the following steps:
+   2. Go to `about:config` and add setting `xpinstall.signatures.required` and set to `false`
 
-	  	```
-	    	rm -rf addon.xpi
-	    	npm run package
-	    	npm run start
-	   	```
+   3. Find the XPI for the environment:
 
-		  This will add an xpi file to the root of your project which you can drag to the browser. You will need to give it permissions when done with this latter method.
+      - Locally: `npm run package` which adds an addon.xpi to the root of the project
 
+      - DEV / STAGE: Visit the dev or stage version of the website and click on "Get Firefox Color" (i.e. open testing.html) and use one of the referenced XPI files
 
-		To debug the background page, you can go to Tools -> Web Developer -> Remote Debugging and click on the "Inspect" button.
+   4. Now load the XPI (from the previous step) to the browser by one of the following ways:
 
-		To toggle the addon on and off or remove, you can go to `about:addons`
+      - Go to `about:debugging` and click on "Load Temporary Add-on..." and add the xpi
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,10 +19,28 @@
...
+
+	  - run the following steps:
+
+	  	```
+	    	rm -rf addon.xpi
+	    	npm run package
+	    	npm run start
+	   	```
+
+		  This will add an xpi file to the root of your project which you can drag to the browser. You will need to give it permissions when done with this latter method.
```
Rob--W, line 35: With the first step, the add-on is installed as a temporary add-on. After making changes to the files on disk, reloading the add-on will result in the browser picking up the changes.

By dragging and dropping the XPI, the add-on will just be updated. There was no need for the `about:debugging` step. If you are able to install an add-on in this way, it is due to the`xpinstall.signatures.required` pref being `false` (which is not the case by default, but done when Firefox is started through the `web-ext` command).

So in a nutshell: The original steps were sufficient for starting the browser with the new extension for the first time. What was missing is that the addon should be reloaded to pick up the changes from the disk.

The xpi-based test flow is also missing. If you want to add that, make it clear that it is either the `about:debugging`-based testing, or `npm run package` + setting the pref (which only works on Firefox Developer Edition or Nightly).

README.md
```diff
@@ -19,10 +19,28 @@
...
 
-1. Visit `http://localhost:8080` to see the web-based theme editor - changes
+	  - In Firefox 57 + open `about:debugging` and load the
+	      `build/extension/manifest.json` file.
+
+	  - run the following steps:
+
+	  	```
+	    	rm -rf addon.xpi
+	    	npm run package
```
Rob--W, line 31: `rm -rf addon.xpi` is not needed, `npm run package` automatically takes care of that.

`npm run start` is not required either. `npm run package` creates the XPI and outputs it to `addon.xpi`.

rebmullin: I thought I had to restart the server but just teseted this, thanks! 👍

README.md
rebmullin: > With the first step, the add-on is installed as a temporary add-on. After making changes to the files on disk, reloading the add-on will result in the browser picking up the changes.

you mean `npm run package` as the first step?

I'm assuming this ^ is what you mean by "first step" 

Regarding `about:debugging`, I left this in b/c it seems easier although I admit that I forgot that a change to the config (xpinstall setting) is required.

so how does something like this sound for all this then ~ : 

....
To load the extension you can do this in the following ways:

LOCALLY:

`npm run package` 

After that script is finished, drag and drop the `addon.xpi` from the root of the project to the browser.  Each time there is change to the background.js file, you will need to re-run `npm run package` and re-drag this to the browser (note: you do not need to restart the server).


DEV and STAGE on Firefox 57 +  Dev or Nightly 

   1. Go to `about:config` and add setting `xpinstall.signatures.required` and set as `false`

   2. Go to the addon's homepage and click on "Get Firefox Color." This will link out to a testing page where you can find the latest unsigned xpis for all the environments.

   3. Next open `about:debugging` and add as temporary add-on


To debug the background page, you can go to Tools -> Web Developer -> Remote Debugging and click on the "Inspect" button.

To toggle the add-on on and off or remove, you can go to `about:addons`

...

Rob--W: What you're describing as "LOCALLY" and "DEV and STAGE" looks the same to me. In both cases, an unsigned add-on is generated and loaded in Firefox. Even if testing "locally", the addon.xpi can only be loaded when the preference is set.

In both cases, `about:debugging` can also be used to load the unsigned xpi files.

Perhaps it's easier to write the documentation if you keep in mind that there are two separate questions, each with two separate steps:

1. Q: How to get a XPI file for the correct environment localhost, dev, stage, prod?
  - A1: Run `npm run package`
  - A2: Visit the dev or stage version of the website and click on "Get Firefox Color" (i.e. open testing.html) and use one of the referenced XPI files.
2. Q: How to load it?
  - A1: about:debugging, temporary add-on.
  - A2: setting the xpinstall.signatures.required pref followed by loading the XPI in the browser (drag and drop, clicking on the link in testing.html, etc.)

rebmullin:
hey @Rob--W, thanks for the suggestions. I just made some updates. Please let me know if it's any better or questions. thanks!

README.md
```diff
@@ -19,10 +19,32 @@
...
+
+      - Locally: `npm run package` which adds an addon.xpi to the root of the project
+
+      - DEV / STAGE: Visit the dev or stage version of the website and click on "Get Firefox Color" (i.e. open testing.html) and use one of the referenced XPI files
+
+   4. Now load the XPI (from the previous step) to the browser by one of the following ways:
+
+      - Go to `about:debugging` and click on "Load Temporary Add-on..." and add the xpi
+
+      - (or) drag and drop the XPI to the browser
```
Rob--W, line 38: Step 2 should be combined with this bullet point. When the "Load Temporary Add-on" button is used, the preference change is not needed.

And also note that this alternative only works on Beta, Developer Edition and Nightly.

rebmullin: @Rob--W, just to make sure i have this right before I change, would be something like this:


1. Find the XPI for the environment:
    * Locally: `npm run package` which adds an `addon.xpi` to the root of the project 
    * DEV / STAGE: Visit the dev or stage version of the website and click on "Get Firefox Color" (i.e. open testing.html) and use one of the referenced XPI files 

2. Now load the XPI (from the previous step) to the browser by one of the following ways:
    * Go to `about:debugging` and click on "Load Temporary Add-on..." and add the xpi 
    * (or) Go to about:config and add setting `xpinstall.signatures.required` and set to `false`. Next drag and drop the XPI to the browser. Note:  you must use Firefox 57 + Beta, Dev or Nightly browser w/ this approach.

Rob--W: Looks good. You can drop the "57+" qualifier from your comments, by the way. Older versions aren't relevant and not worth mentioning.

rebmullin:
@Rob--W , how's it look now?

Issue #942, opened 2020-07-23
Can't set alpha for some colors


andreicristianpetcu:
I have a theme where I set these colors with an alpha.

popup_highlight
sidebar_highlight_text
sidebar_text
toolbar_field_separator
toolbar_field_text_focus
toolbar_top_separator

I can't replicate it with Firefox Color. More context here: https://github.com/mozilla/FirefoxColor/pull/940#issuecomment-662641273

andreicristianpetcu:
I tested all of them and these are the colors that support alpha:

```
export const colorsWithAlpha = [
  "button_background_active",
  "button_background_hover",
  "frame_inactive",
  "frame",
  "icons_attention",
  "icons",
  "ntp_background",
  "ntp_text",
  "popup_border",
  "popup_highlight_text",
  "popup_highlight",
  "popup_text",
  "popup",
  "sidebar_border",
  "sidebar_highlight_text",
  "sidebar_highlight",
  "sidebar_text",
  "tab_line",
  "tab_selected",
  "tab_text",
  "toolbar_bottom_separator",
  "toolbar_field_border_focus",
  "toolbar_field_border",
  "toolbar_field_focus",
  "toolbar_field_highlight_text",
  "toolbar_field_highlight",
  "toolbar_field_separator",
  "toolbar_field_text_focus",
  "toolbar_field_text",
  "toolbar_field",
  "toolbar_text",
  "toolbar_top_separator",
  "toolbar_vertical_separator",
  "toolbar",
];
```

tab_background_separator simply does not work at all on my machine. Not sure if I should add it in the alpha or not.

Rob--W:
What does it mean to not support an alpha?

andreicristianpetcu:
I added all the colors to colorsWithAlpha. I took all of them one by one and made them fully transparent. The ones above took the transparency value. The missing ones stayed opaque regardless of what I did.

andreicristianpetcu:
Also, some if them had a transparency minimum. I could make them transparent but not fully transparent. The popup in the burger menu can be invisible, the megabar can be just partially transparent.

Rob--W:
The Color extension accepted the values though, so I suggest to support alpha for all colors.

andreicristianpetcu:
so shoul I remove colorsWithAlpha?

Rob--W:
It was introduced in #55 (referring to #30), for some reason. I don't know why.

[...]

PR #944, opened 2020-07-26
#942 process colorsWithoutAlpha

andreicristianpetcu:
image


andreicristianpetcu:
Example of a theme using these values (the original classic theme I wanted to port) http://localhost:8080/?theme=XQAAAAJ4BAAAAAAAAABBKYhm849SCicxcUhA3DJozHnOMuotJJIF9jJb-FWmo1SjM2MWLJWW2J5lCZzVP2MQDfAs41MN24hV_0CG0VmlorOZhd96gWR3F52l7wx3zlkkdyKtrV34f6wGULgIRdBYEhR1tpgWvD_P4ZiLWaBEAmloHI_5cxZ9x_cZM_OK88Nl3ZIJe2eS5nyeP8CVtq6NY7LYt67RAbjyabSi8PF-jh1xB-IZixQYUZSZxctWZBDsSopT0DfLlnYr4w6gyAVpf9LjPKmpzX5rhrRh5uLFeKeBveGuNoSVr_IZMKN_fC2jQv7MfdzRvP3RsVULqjnkwGsL2-yCinV5zVwgUSUmTSRh5DTaJWilqNxV1pkqmS-E-dmMAivIUwQLh6V6vdg87cg3Dyu3-gB-sN9zX-Yeb9L8i8PPVDV1jdc4263Vi9iMKS8NE-LhtuFQz_4Lo1Z9Y3ObhijmoTYggwfBNTsMHKMXReSAeh7qpRYNzwg0Nq4ygDrWZRVtOcc6Rxw-o9DlEMMdoAWTpkZghrdIotT1W2Yh_aQXJHDiQy86aHrjUT3mZ0ql56Hc6JUFpfB5-4nGcw

andreicristianpetcu:
Damn.... there is a validator that forces a color to either have or not have alpha for all the themes. What's the point of having `a: 1`?
https://app.circleci.com/pipelines/github/mozilla/FirefoxColor/1015/workflows/1e2bfbc9-ac1a-49bb-ad2c-4bec381a504c/jobs/2707
image

Rob--W:
Interesting, I wasn't aware of `docs/theme-schema.json` before. Could you update it to add all new colors that you added in the advanced colors PR?

And rather than "fixing" the theme defaults by stripping the a:1, change the definition to make "a" optional.

andreicristianpetcu:
done @Rob--W

bin/convert-themes-to-rgb.js
```diff
@@ -1,16 +1,19 @@
...
 
 const makeTinycolor = colorIn => {
   let { a, s } = colorIn;
-  return tinycolor({
+  let newColor = tinycolor({
     ...colorIn,
-    a: Math.floor(a) / 100.0,
     s: Math.floor(s) / 100.0
   });
+  if (a) {
```
Rob--W, line 13: `a != undefined` to make sure that if `a: 0` is used, that it continues to be like that.

And please add a unit test for this case.

src/lib/themes.js
```diff
@@ -125,8 +125,9 @@ export const normalizeThemeBackground = background =>
 
 // Utility to ensure normal & consistent colors
 export const normalizeThemeColor = (name, data, defaultColor) => {
-  const color = makeTinycolor(data || defaultColor).toRgb();
-  if (!colorsWithAlpha.includes(name)) {
+  const inColor = data || defaultColor;
+  const color = makeTinycolor(inColor).toRgb();
+  if (colorsWithoutAlpha.includes(name) || (inColor && !("a" in inColor))) {
```
Rob--W, line 130: If alpha has the default value, it can also be deleted IMO.

Could you add unit tests to cover this new logic?

andreicristianpetcu: I'm a bit confused. I initially proposed this.
image
https://github.com/mozilla/FirefoxColor/pull/944#issuecomment-663971106
Then I implemented this change. https://github.com/mozilla/FirefoxColor/commit/b911023c3d60918866306d3b15b19a80ae7d6b88#diff-28b54e29959f783d6744df0f4e0c3c49R28
image
Then you told me to revert. https://github.com/mozilla/FirefoxColor/pull/944#issuecomment-664003245
image

I'm fine with either way but if we delete the opaque alpha, then the "[should not change preset themes](https://github.com/mozilla/FirefoxColor/blob/410e1eda9f6cc9a4837edefef136398712619c59/src/lib/themes-test.js#L27-L37)" will fail if we don't update the preset themes like I did before.
image

Am I missing some solution where we can delete the `a: 1` but we avoid updating the default themes?

bin/convert-themes-to-rgb.js
andreicristianpetcu: Where is this called from? Should I create a new test for it?

src/lib/themes.js
Rob--W: I thought that you deleted `"a": 1` in order to get it to pass the schema validation.

I'd be fine with removing `"a": 1` from the default theme templates if it is still backwards-compatible.
Could you verify that the website would still be compatible with the old add-on (i.e. the current 2.1.6) if you do remove `"a": 1`?

bin/convert-themes-to-rgb.js
Rob--W: Oh this is not called from anywhere, it's a tool. Never mind the unit test for this file.

You should be able to undo the change though, as the `a` property is deleted below depending on conditions.

src/lib/themes.js
andreicristianpetcu: do we get any benefit if we delete the a: 1?

Rob--W: I suppose that the permalink to the URL would be smaller (i.e. the part after `?theme=...`).

andreicristianpetcu:
@Rob--W here is a URL of the first theme which by default has alpha set to 1 in the toolbar. This URL does not have any alpha. It works fine. https://color.firefox.com/?theme=XQAAAAIwAQAAAAAAAABBqYhm849SCia2CaaEGccwS-xMDPr6QjyuAiEifwJmpSs38LXYwAM0txHvymCyYe8A373Y-5dI6pwZVVCPpCb6CSC650B4-wuqUzhU9zT1DXDh8Cl7n78o7Ucx1t-v-XJHZv9mutHz9tBgdjJSgPDugfulzDxfznqKATnRv7bV9YYnrnpWmkZxS6IwaDDhgLClbtw0iNx1VqSMyOyILuNgYmNImKhHxo2zNo7e43RFZJPqljK0zz_7I6HP6zqBQjyXvPat4nXmJl2M3if_jyCAAA

src/lib/themes.js
```diff
@@ -89,15 +89,16 @@ export const themesEqual = (themeA, themeB) => {
...
-      return false;
-    }
-    if (
-      alphaInA &&
-      alphaInB &&
-      Math.abs(colorA.a - colorB.a) > alphaEqualityTolerance
-    ) {
-      return false;
-    }
+      if ((alphaInA && colorA.a !== 1) || (alphaInB && colorB.a !== 1)) {
```
Rob--W, line 92: The logic of these conditions spread over multiple lines look complicated.

Since absence of `a` is now treated like `a: 1`, use the nullish coalescing operator to get the desired (numeric) value for `alphaA` and `alphaB`. Then you'll only need one if-condition like `if (Math.abs(alphaA - alphaB) > alphaEqualityTolerance) {`

Note that I'm suggesting the nullish coalescing operator instead of OR, to make sure that `0` stays `0` (i.e. transparent) and doesn't become `1` (opaque). Considering these edge cases, it would also be useful to check if there are existing tests for alpha being undefined, 0, 1, in between (e.g. 0.5), and if they don't exist yet, add them.

andreicristianpetcu:
Done @Rob--W :ballot_box_with_check:

src/lib/themes-test.js
```diff
@@ -300,6 +309,33 @@ describe("lib/themes", () => {
...
+    it("should accept if both alpha are set to opaque", () => {
+      expect(
+        subject(
+          { colors: { toolbar: { r: 1, g: 2, b: 3, a: 1 } }, },
+          { colors: { toolbar: { r: 1, g: 2, b: 3, a: 1 } } }
+        )
+      ).to.be.true;
+    });
+
+    it("should accept if both alpha are set to transparent", () => {
```
Rob--W, line 330: Those are "good weather" tests. If you change the implementation below from `?? 1` to `|| 1`, will you see any test failure? From what I have seen in the tests, the answer appears to be no.

To get decent coverage, you should at the very leasy also be comparing `a: 0` with `a: 1` (expect false) and `a: 0` with `(no a)` (expect false).

And since the source code contains the (hardcoded) magic number `1`, it might be a good idea to compare these boundary values 0/1 with a different value in between (e.g. 0.5): compare `a: 0` with `a: 0.5`, and `a: 1` with `a: 0.5`.

andreicristianpetcu:
done @Rob--W :rocket:

andreicristianpetcu:
Did you have time to submit the extension to AMO, @Rob--W ?

Rob--W:
Not yet. I started with some testing last week, found a few issues and fixed them (I just pushed them to master). I haven't had the chance to complete the tests before release.

andreicristianpetcu:
@Rob--W Were you affected by the layoffs? :(

Rob--W:
@andreicristianpetcu Thanks for your concern, I'm not.

On this issue, I was trying to update all dependencies to the latest version (`npm audit fix`), but that broke and I haven't looked back yet.

andreicristianpetcu:
Are you still stuck with the dependencies? Need help? @Rob--W

Rob--W:
:_: Turns out that `npm audit fix` failed because I inadvertently put `package-lock=false` in `.npmrc` somewhere. After removing that `npm audit fix` ran to completion. I've opened a PR at #948 and if nothing fails I'll merge that. I'll need to check that the generated build output (before and after #948) is equivalent and not causing regressions.

andreicristianpetcu:
Any update on the next version in amo @Rob--W ?

andreicristianpetcu:
This got merged 2 months ago. Any ne version in sight? @Rob--W ?

Rob--W:
The dependency update was done in #955, and I have bumped the version to 2.1.7.
The add-on is currently pending review on AMO. I'll notify you when the update has been rolled out.

Rob--W:
The add-on has been approved and I've pushed the latest version of the website to production.

@andreicristianpetcu Thanks again for your contributions and patience.

PR #945, opened 2020-07-31
#942 preview theme

andreicristianpetcu:
https://github.com/mozilla/FirefoxColor/issues/942#issuecomment-667426332

### DIFF
--- a/src/web/lib/components/BrowserPopup/index.scss
+++ b/src/web/lib/components/BrowserPopup/index.scss
@@ -8,7 +8,6 @@
     transform: rotate(45deg);
     position: absolute;
     z-index: -1;
-    background: var(--grey-90-a10);
   }
 
   .browser-popup__caret {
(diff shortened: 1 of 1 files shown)


andreicristianpetcu:
So you want the dashes to have the same color as the popup? When it's transparent, so should the dashes, right?

Rob--W:
There is not an issue with the transparency of the red dashes - them being opaque is fine, as they only appear when the user selects a color. I think that my previous comment wasn't clear, so I'll try to be a bit more detailed below.

The (minor) issue is that there are visible dashes at the intersection of the triangle and the rectangle (from the outline around the rectangle). You can also see this in your screenshot, look near the extrusion at the popup. You'll see dashes (which are painted behind the triangle):

andreicristianpetcu:
I looked into the issue with the border and it seems very complicated. I think it can be done with a really complicated clip path but I'm not sure it's worth my time. If it's ok with you, I say merge it as it is and release it. @Rob--W

Issue #943, opened 2020-07-23
Add option to import theme from zip file


Rob--W:
The "Share" button only supports themes without images, but as soon as an image is included, it is no longer possible to share a linlk.
The next button, "Export", does offer the ability to export the theme, but once exported it cannot be imported in Firefox Color again.

Since we now support all theme properties (#940), it should be possible to offer an Import button that is an accurate representation of the theme.

There is one thing though: a zip file with the theme exension supports arbitrary image formats (including SVG), whereas dynamic themes only support PNG and JPG due to https://bugzilla.mozilla.org/show_bug.cgi?id=1491790 .

Rob--W:
@andreicristianpetcu Is this something that you'd be interested in working on? Fixing this would allow FirefoxColor to more easily import themes with images inside.

andreicristianpetcu:
Sorry but I cannot. I'll focus my free time on about:logins.

himanshujaidka:
I would like to do work on this issue

Rob--W:
To resolve this bug, the following needs to be implemented:

1. Logic that reads a zip file (given by the user), reads the containing manifest.json and convert it to a theme object (its format is specified by https://github.com/mozilla/FirefoxColor/blob/master/docs/theme-schema.json, and examples of valid theme objects can be found at https://github.com/mozilla/FirefoxColor/tree/master/src/preset-themes ). The project already depends on the jszip library, which is currently used to generate a zip file, but it also supports reading zip files (with just a few lines of code). Most of the effort is in converting the data from manifest.json to a theme object plus importing images, if any.
2. Add a UI element (e.g. a button next to the existing Export button) to trigger the logic from step 1.

See the README to get started, by setting up a local development environment.
Let me know if you have any other specific questions.

himanshujaidka:
when I am trying to do npm ci then my system consumes a lot of time

Rob--W:
`npm ci` installs all (dev)Dependencies. It can take several minutes, longer if your internet connection is slow. Just start `npm ci`, take a break and get back.

himanshujaidka:
ok thanks

himanshujaidka:
visual studio is mandatory for local installation? @Rob--W

himanshujaidka:
![Screenshot (47)](https://user-images.githubusercontent.com/58654018/90156822-8f810e80-ddaa-11ea-8466-720237484844.png)

himanshujaidka:
[...]

PR #947, opened 2020-08-17
custom background has shareable link now #943 

himanshujaidka:
and updated to latest version of dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -173,9 +173,9 @@
           }
         },
         "lodash": {
-          "version": "4.17.15",
-          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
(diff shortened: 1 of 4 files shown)


himanshujaidka:
resolves first part of #943

himanshujaidka:
@Rob--W I didn't find any export button.

Rob--W:
The button only appears when the extension has been installed: https://github.com/mozilla/FirefoxColor/blob/f23ab4f66a409c174c06e8d7e9a60f41c5eba0c6/src/web/lib/components/AppHeader/index.js#L156-L162

I don't see a good reason for hiding the button when the extension is not installed. I guess that it was done for visual reasons, while disregarding the loss of functionality, in https://github.com/mozilla/FirefoxColor/commit/b3fa52742d84c39a24a18e949fcd4ff700fbad52 .

Issue #937, opened 2020-07-02
“Terms & Privacy” popup links do not turn purple after being accessed


acornestean:
Preconditions:
1. Go to about:config and change the “xpinstall.signatures.required” pref to “false”
2. Access https://color.stage.mozaws.net/
3. Click on the “Advanced colors” tab
4. Click on “Firefox Color Extension” link from the right side panel
5. On the newly opened page, click on “Unsigned add-on for color.stage.mozaws.net “
6. Follow the instructions on the doorhangers and install the Firefox Color add-on

Steps to reproduce:
1. Access https://color.stage.mozaws.net/
2. Click on the “Privacy” button at the bottom of the page
3. Notice the “Terms & Privacy” popup is displayed
4. Click on the “Websites Privacy Notice” link
5. Return to https://color.stage.mozaws.net/
6. Click on the “Privacy” button at the bottom of the page
7. Notice the “Terms & Privacy” popup is displayed
8. Observe that the “Websites Privacy Notice” link did not turn purple indicating that it has been accessed

NOTE:
The same occurs for the “Websites Terms of Use” link when accessed via the Privacy button popup and both the mentioned links when accessed via the Terms button popup.

Expected results:
The accessed links from the Privacy and Terms popups should become purple as an indication that they have been visited.

Actual results:
The links do not change color to purple after being accessed.

Affected versions:
- Nightly 80.0a1/20200701213212
- Beta Developer Edition 79.0b2/20200630191632
- Release Unbranded 78.0/20200625152958

Affected OSes:
- Windows 10 64-bit
- Ubuntu 16.04 LTS

acornestean:
The issue also occurs on macOS Catalina 10.15 as well.

PR #949, opened 2020-08-25
link color now changes after user visit

himanshujaidka:
#937 issue @Rob--W

### DIFF
--- a/src/web/lib/components/Modal/index.scss
+++ b/src/web/lib/components/Modal/index.scss
@@ -39,9 +39,12 @@
   p {
     margin: 0;
   }
-
-  a:not(.modal__button) {
-    color: var(--blue-50);
+  
+  a{
+  color: blue;
+  }
+  a:visited{
+   color: purple;
   }
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -173,9 +173,9 @@
           }
         },
         "lodash": {
-          "version": "4.17.15",
-          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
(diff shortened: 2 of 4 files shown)


src/web/lib/components/Modal/index.scss
```diff
@@ -39,9 +39,12 @@
   p {
     margin: 0;
   }
-
-  a:not(.modal__button) {
```
Rob--W, line 43: Wouldn't just using `a:not(.modal__button):not(:visited)` have worked?

himanshujaidka: a:not(.modal__button):not(:visited){
    color:var(--blue-50);
  }
it will not work @Rob--W

Rob--W: That's interesting, because if I modify the CSS rule via the devtools of the browser, it does work: the visited link becomes purple.

How did you verify that "it will not work"?

himanshujaidka: Screenshot (56)
@Rob--W actually this is confusing me, It turns to purple color but I am not sure that after visiting it will turn to purple or at a first look it is a purple

Rob--W: You can delete the entry from the browser history to see its normal color. Or open it in private browsing mode.

himanshujaidka: a:not(.modal__button):not(:visited) {
    color: var(--blue-50);
  }
is I am writing correct code?

Rob--W: Yes.

himanshujaidka: then I will make pr of this code :)

Issue #883, opened 2020-01-29
Style: the pop-up styling ui looks too intrusive


co16353sidak:
I love this new tool with which I can customise my firefox instance using my colour preferences and et al.
A certain issue that I came across is that the popup UI looks too intrusive to me, the way it is designed it appears intrusive and the user ends up clicking on the pop-up button thinking it will pop back in.
This could cause problems for people (like me) who have OCD or issues with a pop-up obstructing my field of view.
I have attached the screenshot below:

![Obtrusive pop-up](https://user-images.githubusercontent.com/35738138/73328170-c2d0d580-427e-11ea-90d6-aec8e938227e.png)
As you can see above that the pop-up seems to block the colour-picker which I feel is obtrusive as well as an example of bad UX design
In my opinion, there should either be an option to close the pop-up or a trivial fix could be to just shift the colour picker down so that it is no longer covered by the aforementioned pop-up,
would also love to work on this issue if it is possible as a first-time contributor here. :1st_place_medal:

Rob--W:
Hi @co16353sidak, thanks for your report. I think that it we should either shrink the panel or increase the minimum size of the gap between the "location bar" and the theme generator below.

If you're interested, I'm willing to mentor you on the implementation of the fix.

co16353sidak:
@Rob--W thank you for the response,
Would love to work on this, please advise on which files to start working on so that i can send a PR with the changes asap. :tada: :100:

Rob--W:
To find what you need to change, use the browser's devtools to inspect the document. Then you can find elements (and their class names) of interest.

The popup is only visible when the extension is installed: https://github.com/mozilla/FirefoxColor/blob/31019674c050dd46ccaf6889108a408e1fa55328/src/web/lib/components/App/index.js#L129

When the extension is installed, the installation banner is not shown, but a dummy element is shown instead: https://github.com/mozilla/FirefoxColor/blob/31019674c050dd46ccaf6889108a408e1fa55328/src/web/lib/components/App/index.js#L142

Its styles are declared here: https://github.com/mozilla/FirefoxColor/blob/c27fb6849a0323ddc2af3c66e0993b16b80b1d00/src/web/lib/components/App/index.scss#L21

Je-ni:
hello, @co16353sidak are you still working on this? I'd love to work on it if you're not.

GeekBoySupreme:
Would love to help out with this if you need any help @co16353sidak

Rob--W:
[...]

PR #962, opened 2021-01-04
fix: styling of popup

himanshujaidka:
fix the issue no #883 and update the dependencies or packages
Screenshot (122)


Rob--W (changes requested):
This may work for your specific screen size, but this hardcoded value of 22px is not the universally right fix. For example, shrink the window size (reduce the height of the browser window) and you'll see that there is still some overlap.

And also, please don't put unrelated changes (e.g. dependency updates) together with this patch. It increases the size of the patch and increases the risk of merge conflicts.

himanshujaidka:
> This may work for your specific screen size, but this hardcoded value of 22px is not the universally right fix. For example, shrink the window size (reduce the height of the browser window) and you'll see that there is still some overlap.
> 
> And also, please don't put unrelated changes (e.g. dependency updates) together with this patch. It increases the size of the patch and increases the risk of merge conflicts.

okay

himanshujaidka:
> This may work for your specific screen size, but this hardcoded value of 22px is not the universally right fix. For example, shrink the window size (reduce the height of the browser window) and you'll see that there is still some overlap.
> 
> And also, please don't put unrelated changes (e.g. dependency updates) together with this patch. It increases the size of the patch and increases the risk of merge conflicts.

It seems to be working fine even in the small shrunk tab or in the mobile view

himanshujaidka:
Made changes

src/web/lib/components/App/index.scss
```diff
@@ -24,7 +24,11 @@
     background-position: center center;
     background-repeat: no-repeat;
     background-size: 164px 164px;
-    padding-bottom: 22px;
     flex: 1;
+
+    
+    @media only screen and (max-width: 2500px){
```
Rob--W, line 30: What's up with this arbitrary 2500px maximum width?

If anything, I would expect the condition to be based on the height.

Rob--W (changes requested):

src/web/lib/components/App/index.scss
himanshujaidka: okay

himanshujaidka: pls have a review

himanshujaidka:
sir, any updates?

src/web/lib/components/App/index.scss
```diff
@@ -24,6 +24,7 @@
     background-position: center center;
     background-repeat: no-repeat;
     background-size: 164px 164px;
+    min-height: 28px;
```
Rob--W, line 27: Can you explain why you've chosen 28px? First you chose 22px, now 28px. I'd like to understand the reasoning behind it.

Rob--W (commented):
Please undo the update of the `package.json` / `package-lock.json` files. These are unrelated to the patch.

src/web/lib/components/App/index.scss
himanshujaidka: because when I am running it on my local system, 22 px is not a perfect fit, somewhere it is collapsed with the banner but 28 px is a perfect fit on my screen, so, therefore, I have chosen this 28px

himanshujaidka:
> Please undo the update of the `package.json` / `package-lock.json` files. These are unrelated to the patch.

okay, I will do that sorry that I have forgotten to remove it as you have said earlier also

Rob--W:
Closing in favor of #966, which solves the issue in a better way.

himanshujaidka:
> Closing in favor of #966, which solves the issue in a better way.

Thanks for the guidance :)

PR #966, opened 2021-02-01
Fixes #883: set .browser min-height: 700px

### LINKED ISSUE #883: Style: the pop-up styling ui looks too intrusive

### LINKED ISSUE #952: Saved themes not accessible
The fake menu is overlaying the "Saved themes" Tab

Screenshot_2020-09-05 Firefox Color

Screenshot_2020-09-05 Firefox Color(1)

Saffafatima12:
Fixes #883 
Fixes #952

I have increased the 'min-height' property for '.browser' in 'src/web/lib/components/App/index.scss' from 600px to 700px. This ensures that the popup never overlaps with the 'Raw Theme Data' bar or the color selector below. 

Before edit:

600px

After edit:

700px

### DIFF
--- a/src/web/lib/components/App/index.scss
+++ b/src/web/lib/components/App/index.scss
@@ -15,7 +15,7 @@
 
   > .browser {
     grid-column: 2 / 3;
-    min-height: 600px;
+    min-height: 700px;
   }
 
   .app__firefox {


Rob--W:
Looks good, thanks for the patch, explanation and the screenshots. I locally verified that this produces the desired result.

The metadata of the commit is `root <root@DESKTOP-EM8PF0U.localdomain>`, which doesn't look entirely right. Do you want to fix it up before I merge the patch?

To configure your username/email (in `.gitconfig`) and/or on Github, see https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address

After editing your user name, you can use `git commit --amend --reset-author` to update your patch and force-push the updated commit to your branch to get the update here.

Saffafatima12:
Thank you so much for checking the patch. I apologize I did not check the metadata of the commit. It seems to be the name of my root directory in WSL. I hope I have correctly updated my username and email now. Thanks for pointing me to the GitHub documentation and please do let me know if there are still some issues that need to be corrected.

Rob--W:
The e-mail is now set correctly, but the name is still "root" rather than your name. You can check how it looks like by running `git log` (or use `git log -n1` to show only the last entry). That will show the author, date and commit message. This is what it looks like right now:

```
$ git log -n1
commit 1679491e4700fc2a3fac4571e8032569179148ca
Author: root <HIDDEN-FOR-PRIVACY@gmail.com>
Date:   Fri Feb 5 09:50:29 2021 +0500

    Fixes:#883 set .browser min-height: 700px
```

To set the name, follow the similar instructions as before, but with `git config user.name` instead of `git config user.email`.

Saffafatima12:
I get this with git log -n1 now:

Author: saffafatima <saffafatima12@gmail.com>
Date:   Sun Feb 7 10:56:28 2021 +0500

    Fixes:#883 set .browser min-height: 700px

Rob--W:
Thanks!

PR #975, opened 2021-04-27
Explain how the exported theme can be used

### LINKED ISSUE #971: Cannot export theme, extension does not work
***

# Cannot export theme

This has been a common issue I have noted by looking at reviews from the past 7 months. I went to install this extension today, and I found that every time I exported the file, it said that it couldn't install the extension, as the extension was corrupt. I can confirm it is, as there is a <600 byte blank SVG file, and a normal JSON file. I am surprised that this isn't being reported as much as it should. I looked through issues before reporting, but couldn't find any recent ones mentioning this, so I will say it here. This may be a duplicate, as this apparently has been an issue for 7 months or more.

I tried to include a screenshot, but the error message goes away before my system can get a screenshot (using F10: PRT SC/Print Screen)

## Device specs

Browser version: Firefox Browser 87.0 for Ubuntu Linux 20.04.1
Operating system: Ubuntu Linux 20.04.1
DE: GNOME 3.36.1

***

Rob--W:
Fixes #971.

### DIFF
--- a/src/web/lib/components/ExportThemeDialog/index.js
+++ b/src/web/lib/components/ExportThemeDialog/index.js
@@ -76,6 +76,11 @@ class ExportThemeDialog extends React.Component {
     }
 
     if (shouldOfferExportedTheme) {
+      // exportedTheme has XPI MIME set in src/web/lib/export.js
+      const XPI_MIME = "application/x-xpinstall";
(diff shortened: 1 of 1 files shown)


caitmuenster (approved):

caitmuenster:
Thank you, @Rob--W! This WFM.

PR #976, opened 2021-04-27
fix: mark advanced colors that are gone in Proton

### LINKED ISSUE #969: Updates to themeable properties with Firefox 89
We should update Firefox Color to reflect changes coming to themes for Firefox 89 (release date: June 1, 2021): 

* `tab_background_separator` will no longer be supported
* `toolbar_field_separator` will no longer be supported

We'll need to remove those properties from the Advanced Colors section of Firefox Color.

Rob--W:
Fixes #969

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -49,7 +49,7 @@ export const advancedColorLabels = {
   sidebar_highlight: "Sidebar Highlight",
   sidebar_text: "Sidebar Text",
   sidebar: "Sidebar",
-  tab_background_separator: "Tab Background Separator",
+  tab_background_separator: "Tab Background Separator (FF88-)",
   tab_loading: "Tab Loading",
   tab_selected: "Tab Selected",
   tab_text: "Tab Text",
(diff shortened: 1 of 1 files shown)


caitmuenster:
@Rob--W, I wasn't clear from my local test where this is should show up (if it supposed to show up in the UI at all). Could you give a quick overview of how this change is applied?

Rob--W:
The "Advanced colors" tab should show this.

caitmuenster:
I see it! 
Screen Shot 2021-04-28 at 2 00 14 PM

I'm not sure it would be understandable to non-technical theme artists, though. How would you feel about putting an asterisk by the name of the property instead and have the X link to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme#colors? We should have an update coming in soon that will mark `tab_background_separator` and `toolbar_field_separator` as unsupported in 89+.

Rob--W:
> I'm not sure it would be understandable to non-technical theme artists, though. How would you feel about putting an asterisk by the name of the property instead

I tried to choose a suffix that was short enough to not overflow the field. An asterisk would be sufficiently small too, but not very obvious as readers would have to figure out where the asterisk is explained. I can turn it into a "deprecated" icon that explains what happens in a tooltip.

> and have the X link to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme#colors?

That link already exists in the side panel.

> We should have an update coming in soon that will mark `tab_background_separator` and `toolbar_field_separator` as unsupported in 89+.

MDN currently shows an image that explains the role of all (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme/themes_components_annotations.png (permalink to [source on Github](https://github.com/mdn/content/blob/3e5d8947cf03d1ae758aff8926c04c38eca2db2a/files/en-us/mozilla/add-ons/webextensions/manifest.json/theme/themes_components_annotations.png))). I hope that both are kept, so that it may be easier for theme devs to migrate. Possibly by taking the existing image and highlighting the balloons of the color names (e.g. by background/border color).

PR #1000, opened 2022-02-02
chore: Bump minimum supported Node.js version to 14

Rob--W:
CI is currently broken because the minimum Node version is too old.

Merge this if CI is green.


Rob--W:
The PR is red because the images in the circleci config need to be bumped too, see https://github.com/mozilla/web-ext/issues/2371

Rob--W:
It's green :)

PR #1054, opened 2023-06-23
Minor readme updates

martinthomson:
Node 14, commas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## Get Started
 
-1. Install Node 10.18.1+ (e.g. using [node version manger][nvm])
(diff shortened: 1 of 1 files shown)


Rob--W (approved):
I bumped the minimum version to Node 14 in #1000.

There is no universal consensus on whether `i.e.` should be followed by a comma (https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e). Personally I usually use `i.e.` without comma after it, but I don't mind merging this PR that adds commas, because the bump in suggested Node version is correct.

